PR #6, opened 2022-12-30
Add new icons

nshandra:
### :pushpin: References

-   **Issue:**  [[Home Page app] Design icon & pages style](https://app.clickup.com/t/2k5e7pv)

### :memo: Implementation
- Changed default icons from "dhis2-app-skeleton" to new ones.

### :art: Screenshots
appHub
image
favicon
image
Icon small is only intended when the image will be displayed in small sizes like the favicon.


adrianq:
@nshandra @ifoche asked me to review the feedback received from the University of Oslo. They have complained about the app not having a proper logo. This should be fixed with this PR. 
There are two more things I think we can take care of as part of this PR:
- Remove "DHIS2" from the name of the app
- They said: "Also looks like the default icons in the App does not work when it's installed. Looks like the base-url is wrong, (drop the "/" in the img-src). " Can you please have a look at it?
I also noticed you move the PR back to draft. What is missing here?

adrianq:
@nshandra as you are a bit busy right now, I will ask @dumebi to perform the above changes. Do not worry for now and give me a shout if you have some bandwidth

p3rcypj:
@adrianq The PR was moved back to draft because the small icon was missing. In the previous favicon, the "waves" and the pencil whitespace, were creating some distortions when the displaying size is very small. So I joined the waves and removed that whitespace. Here some practical examples: [This is how some logos change their design depending on the size to which they are applied](https://brandemia.org/asi-es-como-algunos-logos-cambian-su-diseno-segun-el-tamano-al-que-se-aplican)

p3rcypj:
We were also using favicon.ico that doesn't allow icons with sizes upper than 255x255. So now we use .png icon files separately. Min resolution generated is 192x192, while the app icon on dhis hub will use a 1024x1024 icon (22.5KB).

p3rcypj:
I guess not but if the icons will be used on an installable web app please let me know to generate the icons for each platform (windows, android, ios)

adrianq:
@p3rcypj Is there a reason to use a different blue for this icon? See: 
image
cc @ifoche

adrianq:
Thanks @p3rcypj for the quick fix! it looks good!

PR #7, opened 2023-01-20
homepage fixes

deeonwuli:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/865bhqywt


package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "dhis2-home-page",
```
adrianq: Why are you keeping "dhis2" in the name of the app? The zip file generated during the build process is named "dhis2-home-page" which may be confusing for the user

adrianq (changes requested):
@deeonwuli Just one comment below. Btw, why are the branch and the PR name as new-icons? This has nothing to do with icons, right?

deeonwuli:
> @deeonwuli Just one comment below. Btw, why are the branch and the PR name as new-icons? This has nothing to do with icons, right?

that was a mistake on my part. i've renamed the PR now though.

src/webapp/pages/about/AboutPage.tsx
```diff
@@ -11,7 +11,7 @@ export const AboutPage = () => {
     const contents = [
         `# ${i18n.t("About Home Page App")}`,
         `#### ${i18n.t("Distributed under GNU GLPv3")}`,
-        i18n.t("DHIS2 Home page App is a DHIS2 application that aims to provide direct links to Dhis2 applications."),
```
tokland, line 14: [very minor] `DHIS2`, then `Dhis2`.

tokland (approved):

PR #9, opened 2023-02-14
fix: scroll to top on every render

deeonwuli:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/860pwnprq


ifoche (approved):
thanks @deeonwuli

PR #11, opened 2023-03-01
Feat/allow multiple landing pages

deeonwuli:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/860pyu9px

### :art: Screenshots
Screenshot 2023-03-01 at 16 02 36
Screenshot 2023-03-01 at 15 45 23
Screenshot 2023-03-01 at 16 01 41


ifoche (changes requested):
@deeonwuli I cannot make it work. I am compiling and doing a yarn start in this branch, but I don't get the app. Instead what I get is that I'm redirected to the dashboard, and pointing to a strange URL:
screenshots_1678884674_1351_15032023_744x611
screenshots_1678885026_1357_15032023_614x120

Could you please have a look? I'm just using the dev.user in our instance. Didn't do any special configuration but adding the Homepage app user role to the dev.user

ifoche (changes requested):
@deeonwuli could you place make it that when you have only one landing page, that one is automatically opened?

ifoche (approved):
@deeonwuli, some comments here: 
- there is at least a couple of times that after editing a landing page and saving, the landing page becomes a single element instead of an array in the datastore 
- it looks like the landing page only appears if you share it with the user, but not if you add a group the user is in

I'm mergin this code instead of waiting for that code because we need it by Monday, and I need the landing pages in place to start doing some surrounding work, but please, if you could work in a separate PR to fix them, that would really be helpful. Thank you!

PR #17, opened 2023-04-04
Fix: usergroups sharing settings

deeonwuli:
### :pushpin: References

-   **Issue:** Landing pages are not shared with the user groups in sharing settings


ifoche (approved):
thanks @deeonwuli

PR #19, opened 2023-04-27
Extend sharing settings for sec, subsecs and cats, fix sharing settings in actions and render landing pages in a single tier 

deeonwuli:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/865c67vh7

### DIFF
--- a/i18n/en.pot
+++ b/i18n/en.pot
@@ -5,8 +5,8 @@ msgstr ""
 "Content-Type: text/plain; charset=utf-8\n"
 "Content-Transfer-Encoding: 8bit\n"
 "Plural-Forms: nplurals=2; plural=(n != 1)\n"
-"POT-Creation-Date: 2023-04-06T12:33:26.549Z\n"
-"PO-Revision-Date: 2023-04-06T12:33:26.549Z\n"
+"POT-Creation-Date: 2023-04-28T15:57:52.436Z\n"
+"PO-Revision-Date: 2023-04-28T15:57:52.436Z\n"
 
 msgid "Field {{field}} cannot be blank"
 msgstr ""
@@ -198,6 +198,24 @@ msgid "Field name is mandatory"
 msgstr ""
 
 msgid "Title"
+msgstr ""
+
+msgid "Icon Location"
+msgstr ""
+
+msgid "Top"
+msgstr ""
+
+msgid "Bottom"
+msgstr ""
+
+msgid "Page Rendering"
+msgstr ""
+
+msgid "Multiple Page"
+msgstr ""
+
+msgid "Single page"
 msgstr ""
 
 msgid "Actions"
--- a/src/data/entities/PersistedConfig.ts
+++ b/src/data/entities/PersistedConfig.ts
@@ -2,6 +2,7 @@ import { Permission } from "../../domain/entities/Permission";
 
 export interface LandingPagePermission extends Permission {
     id: string;
+    publicAccess: string;
 }
 
 export interface PersistedConfig {
--- a/src/data/entities/PersistedConfig.ts
+++ b/src/data/entities/PersistedConfig.ts
@@ -1,9 +1,4 @@
-import { Permission } from "../../domain/entities/Permission";
-
-export interface LandingPagePermission extends Permission {
-    id: string;
-    publicAccess: string;
-}
+import { LandingPagePermission, Permission } from "../../domain/entities/Permission";
 
 export interface PersistedConfig {
     poeditorToken?: string;
--- a/src/data/repositories/Dhis2ConfigRepository.ts
+++ b/src/data/repositories/Dhis2ConfigRepository.ts
@@ -1,4 +1,4 @@
-import { Permission } from "../../domain/entities/Permission";
+import { LandingPagePermission, Permission } from "../../domain/entities/Permission";
 import { ConfigRepository } from "../../domain/repositories/ConfigRepository";
 import { D2Api } from "../../types/d2-api";
 import { cache } from "../../utils/cache";
@@ -6,7 +6,7 @@ import { DataStoreStorageClient } from "../clients/storage/DataStoreStorageClien
 import { Namespaces } from "../clients/storage/Namespaces";
 import { StorageClient } from "../clients/storage/StorageClient";
 import { Instance } from "../entities/Instance";
-import { LandingPagePermission, PersistedConfig } from "../entities/PersistedConfig";
+import { PersistedConfig } from "../entities/PersistedConfig";
 import { getD2APiFromInstance, getMajorVersion } from "../../utils/d2-api";
 import { User } from "../../domain/entities/User";
 import _ from "lodash";
--- a/src/data/repositories/Dhis2ConfigRepository.ts
+++ b/src/data/repositories/Dhis2ConfigRepository.ts
@@ -9,6 +9,7 @@ import { Instance } from "../entities/Instance";
 import { LandingPagePermission, PersistedConfig } from "../entities/PersistedConfig";
 import { getD2APiFromInstance, getMajorVersion } from "../../utils/d2-api";
 import { User } from "../../domain/entities/User";
+import _ from "lodash";
 
 export class Dhis2ConfigRepository implements ConfigRepository {
     private instance: Instance;
@@ -97,7 +98,18 @@ export class Dhis2ConfigRepository implements ConfigRepository {
         const config = await this.getConfig();
         const landingPagesPermissions = config.landingPagePermissions ?? [];
 
-        return landingPagesPermissions;
+        const persisted = (await this.storageClient.getObject<any[]>(Namespaces.LANDING_PAGES)) ?? [];
+
+        const rootId: string =
+            _.every(persisted, node => Array.isArray(node)) && !_.isEmpty(persisted)
+                ? _.flatten(persisted)[0].id
+                : _.isEmpty(persisted)
+                ? persisted[0]?.id
+                : "";
+
+        return _.isEmpty(landingPagesPermissions)
+            ? [{ id: rootId, publicAccess: "r-------", userGroups: [], users: [] }]
+            : landingPagesPermissions;
     }
 
     public async updateLandingPagePermissions(update: Partial<LandingPagePermission>, id: string): Promise<void> {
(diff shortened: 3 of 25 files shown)


i18n/en.pot
```diff
@@ -5,8 +5,8 @@ msgstr ""
 "Content-Type: text/plain; charset=utf-8\n"
```
tokland, line 5: `yarn lint` -> 7 problems (0 errors, 7 warnings). Even if not involved with the current feature, it's better to fix them so the branch stays clean and you see what problems the new code added (if any).

src/data/entities/PersistedConfig.ts
```diff
@@ -2,6 +2,7 @@ import { Permission } from "../../domain/entities/Permission";
 
 export interface LandingPagePermission extends Permission {
```
tokland: This type is used in the domain and presentation, so it cannot be a data entity, it looks like a domain entity.

src/data/entities/PersistedConfig.ts
```diff
@@ -2,6 +2,7 @@ import { Permission } from "../../domain/entities/Permission";
...
     id: string;
+    publicAccess: string;
```
tokland: I guess we use the DHIS2 permission string for simplicity (i.e. it's being used in some Sharing component that understands it), but let's add a comment explaining the format of the string.

src/data/repositories/Dhis2ConfigRepository.ts
```diff
@@ -97,26 +98,42 @@ export class Dhis2ConfigRepository implements ConfigRepository {
         const config = await this.getConfig();
         const landingPagesPermissions = config.landingPagePermissions ?? [];
 
-        return landingPagesPermissions;
+        const persisted = (await this.storageClient.getObject<any[]>(Namespaces.LANDING_PAGES)) ?? [];
```
tokland: We should not use the type `any`

src/data/repositories/Dhis2ConfigRepository.ts
```diff
@@ -97,26 +98,42 @@ export class Dhis2ConfigRepository implements ConfigRepository {
...
+
+        const rootId: string =
+            _.every(persisted, node => Array.isArray(node)) && !_.isEmpty(persisted)
```
tokland: Not sure of the type of `node` (because of the `any`), but being the types decided by us (I assume), we should never need checks like `isArray`, which is a signal the type is not right.

src/data/repositories/LandingNodeDefaultRepository.ts
```diff
@@ -252,7 +256,7 @@ const updateLandingNode = (
     } else {
         const newLandingNode = models.map(model => {
             const landingNode = model.find(model => model.id === rootItem?.parent);
-            if (landingNode) {
+            if (!landingNode) {
                 model.push(...items);
                 return model;
```
tokland: This is old code, but remember that variables must be kept inmutable, no `push`. `return [...model, ...items]` or `return _.concat(model, items)`. If we really need to update `model` in-place, then something went wrong.

src/domain/entities/LandingNode.ts
```diff
@@ -16,6 +16,7 @@ export interface LandingNode {
     type: LandingNodeType;
     icon: string;
     iconLocation: string;
+    pageRendering: string;
```
tokland: Isn't this property limited to some values? (single, multiple), then we need a union. Similar to property `type`.

src/webapp/components/additional-components/AdditionalComponents.tsx
```diff
@@ -6,18 +6,38 @@ import { useAppContext } from "../../contexts/app-context";
...
 
-    const rowSize = actions.length === 3 ? 3 : 4;
+                      const hasUserAccess = actionUsers.includes(user.id);
+                      const hasUserGroupAccess = _.intersection(actionUserGroups, userGroupIds).length > 0;
+                      const hasPublicAccess = action.publicAccess !== "--------";
+
+                      return hasUserAccess || hasUserGroupAccess || hasPublicAccess;
+                  })
+                  .map(({ id }) => id)
+            : [];
```
tokland: The ternary operator should be used for short lines (ideally, single-line, at most 3 -> condition, value1, value2). Double ternary operators should be almost always avoided, and instead write a full-fledged if/else if / else if / else block (usually moved to a function, so we can use `return` in each branch)

src/webapp/components/additional-components/AdditionalComponents.tsx
```diff
@@ -6,18 +6,38 @@ import { useAppContext } from "../../contexts/app-context";
...
     isRoot: boolean;
     currentPage: LandingNode;
 }> = ({ isRoot, currentPage }) => {
     const { actions, translate, launchAppBaseUrl } = useAppContext();
 
-    const { showAllActions } = useConfig();
+    const { showAllActions, user } = useConfig();
 
-    const pageActions = isRoot && showAllActions ? actions.map(({ id }) => id) : currentPage?.actions ?? [];
+    const currentPageActions = actions.filter(action => currentPage.actions.includes(action.id)) ?? [];
```
tokland: Many `?? [` in the code are unnecessary according to the typing, as they are non-optional arrays.

src/webapp/components/additional-components/AdditionalComponents.tsx
tokland: Same about ternary operators, but also, this kind of logic code should be in the domain/entity, that's way too complex for a code in a presentation component and not directly related to view logic.

src/webapp/components/item-root/ItemRoot.tsx
```diff
@@ -29,22 +29,34 @@ export const ItemRoot: React.FC<{
...
+                            key={`node-${node.id}`}
+                            isRoot={isRoot}
+                            openPage={() => openPage(node)}
+                            currentPage={node}
+                        />
+                    ))
+                ) : (
+                    <Cardboard rowSize={4} key={`group-${currentPage.id}`}>
+                        {currentPage.children.map((item, idx) => {
+                            return (
```
tokland: [minor] In a map in a component, use arrow style without return (which reduces one indentation level)

src/webapp/components/landing-page-edit-dialog/LandingPageEditDialog.tsx
```diff
@@ -41,6 +42,7 @@ export const LandingPageEditDialog: React.FC<LandingPageEditDialogProps> = props
 
     const [value, setValue] = useState<LandingNode>(initialNode ?? buildDefaultNode(type, parent, order));
     const [iconLocation, setIconLocation] = React.useState(value.iconLocation === "bottom" ?? false);
+    const [pageRendering, setPageRendering] = React.useState(value.pageRendering === "single" ?? false);
```
tokland: [minor] `a === b ?? false` -> `a === b`

src/webapp/pages/home/HomePage.tsx
```diff
@@ -139,6 +131,31 @@ export const HomePage: React.FC = React.memo(() => {
     );
 });
 
+function updateLandingNodes(nodes: LandingNode[], permissions: LandingPagePermission[], user: User): LandingNode[] {
```
tokland: Move logic to some entity.

src/webapp/pages/home/HomePage.tsx
```diff
@@ -139,6 +131,31 @@ export const HomePage: React.FC = React.memo(() => {
...
+    return _(nodes)
+        .map(node => {
+            const pagePermission = permissions?.find(permission => permission.id === node.id);
+            if (!user) return null;
```
tokland: `user: User`, why the check?

src/data/entities/PersistedConfig.ts
deeonwuli: would the comment look something like:
`publicAccess is '--------' | 'r-------'  | 'rw------'`?

src/webapp/pages/home/HomePage.tsx
deeonwuli: this was incase the user is undefined, but i see now that it's unnecessary

src/data/repositories/LandingNodeDefaultRepository.ts
```diff
@@ -250,15 +250,12 @@ const updateLandingNode = (
 
     if (isItemSavedInDatastore)
         return models.map(model => model.map(persisted => (persisted.id === items[0]?.id ? items[0] : persisted)));
-    else if (importNewNode) {
-        models.push(items);
-        return models;
-    } else {
+    else if (importNewNode) return _.concat(models, items);
```
tokland: Type of `models` and `items` are not equal. The equivalent to the old code is: `_.concat(models, [items])`. That's if the old code was correct, of course.

src/data/repositories/LandingNodeDefaultRepository.ts
```diff
@@ -250,15 +250,12 @@ const updateLandingNode = (
...
+    else {
         const newLandingNode = models.map(model => {
             const landingNode = model.find(model => model.id === rootItem?.parent);
             if (!landingNode) {
-                model.push(...items);
-                return model;
+                return [...model, ...items];
```
tokland: Naming is very important. The var is called `model` but the type is `PersistedLandingNode[]`. As `models` hold the array of arrays, the name for that can be `modelGroup`, anything that indicates a plural.

src/domain/entities/Action.ts
```diff
@@ -87,3 +89,28 @@ export const defaultAction: PartialAction = {
...
+                const userGroupIds = user.userGroups?.map(userGroup => userGroup.id);
+
+                const hasUserAccess = actionUsers.includes(user.id);
+                const hasUserGroupAccess = _.intersection(actionUserGroups, userGroupIds).length > 0;
+                const hasPublicAccess = action.publicAccess !== "--------";
+
+                return hasUserAccess || hasUserGroupAccess || hasPublicAccess;
+            })
+            .map(({ id }) => id);
+    else return [];
```
tokland: [minor, subjective] Unless for some specific reason,  prefer if/else if/else blocks to have `{}` brackets to improve readability.

src/domain/entities/Permission.ts
```diff
@@ -4,3 +4,9 @@ export interface Permission {
     users?: NamedRef[];
     userGroups?: NamedRef[];
 }
+
+export interface LandingPagePermission extends Permission {
+    id: string;
+    // publicAccess is '--------' | 'r-------'  | 'rw------'
+    publicAccess: string;
+}
```
tokland, line 11: Proposal `publicAccess: string; // "--------" | "r-------"  | "rw------"`

tokland (approved):

PR #24, opened 2023-08-21
[refactor] Show progress on app load (no flickering) when redirecting

tokland:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/865cqytpg

### :memo: Implementation

- Use a boolean state to signal if a redirection is currently active

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
     "name": "home-page",
     "description": "Home Page App",
-    "version": "1.6.0",
+    "version": "1.6.1",
     "license": "GPL-3.0",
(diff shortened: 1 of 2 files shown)


adrianq:
@tokland Just one minor thing, can we add a generic message saying something like "Loading the user configuration..."?

tokland:
- [x] Add message "Loading the user configuration" for initial loading.

Note that after 8secs, the flag "isLoadingLong" is set and the message changes to "First load can take a couple of minutes, please wait..."

PR #26, opened 2023-12-11
fix: app name rolled back to its previous situation

ifoche:
### Bugfixing
- App name rolled back to avoid conflict with former landing page


MiquelAdell (approved):
I'm not sure why review pad asked my review, if it was automated at all. In any case I like Homepage App more than Home Page so I'm approving it.

PR #27, opened 2024-02-12
feat: send pageview to google analytics 4 when clicking action

p3rcypj:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/8693u5ak9

### :memo: Implementation

On click actions, it will register the URL of the destination, and the name of the action as the title of the pageview.
Also, navigating to other landings and sections will send a pageview as expected.

### :art: Screenshots

- Test GA4 code: `G-EDDE0H1YQ1`

image

- Pageviews

image

- `page_location` param

image
image

### :fire: Testing

Add the `G-EDDE0H1YQ1` GA4 code under the settings page.
The docker I've been using was the CPR one, but you can use whatever u prefer.


anagperal (changes requested):
thanks @p3rcypj ! Just some comments:

- Instead of using directly compositionRoot.analytics.sendPageView in HomePage and in AdditionalComponents, create an useAnalystics hook to encapsulate this so it can be used throughout the app
- Also in https://github.com/EyeSeeTea/uhcpw we use https://www.npmjs.com/package/react-ga4. Maybe for the future is easier to config and use?

src/webapp/pages/settings/useConfig.ts
```diff
@@ -28,6 +31,14 @@ export function useConfig(): useConfigPloc {
         [compositionRoot]
     );
 
+    const updateGoogleAnalyticsCode = useCallback(
+        async (code: string) => {
+            setGoogleAnalyticsCode(code);
+            compositionRoot.config.updateGoogleAnalyticsCode(code);
```
anagperal: Add await here, if not in SettingsPage component, when you are saving the G4 code, the page reloads without waiting to save the code, so never is saved

MiquelAdell (commented):
Finally got to test this.
I identified some functionality that may be missing or not needed.
I can confirm that views are tracked correctly if "the destination page" is inside home page app. If it is outside of it (it goes somewhere else) it's not tracked.

I don't know if we need to track those links.

If we need to track those links we could capture the click, send the event, and then resume the click

I recorded a video but I think it may not be needed if the text explanation is clear enough:

https://github.com/EyeSeeTea/home-page-app-dev/assets/4996457/c8879d58-3985-4f27-aec5-87bd8674ac62

p3rcypj:
Just a comment for @MiquelAdell and @anagperal, events on real-time may take 20 seconds or even 1 minute to show in the report (yes "real-time", google should add small info about this)

anagperal (approved):
thanks @p3rcypj ! I've checked [commit](https://github.com/EyeSeeTea/home-page-app-dev/pull/27/commits/fdfd91b2ee3a6212ceb6c15b91d9217e1e3d99f7) and [commit](https://github.com/EyeSeeTea/home-page-app-dev/pull/27/commits/e6a472a65a773dbc59d7aa6fa509227637ffcc1a)  and code it's ok

PR #28, opened 2024-02-12
fix: return correct children when building landing node

deeonwuli:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/8693refcm

### :fire: Testing
`REACT_APP_DHIS2_BASE_URL=https://dev.eyeseetea.com/who-dev-238/`

### DIFF
--- a/src/data/repositories/LandingNodeDefaultRepository.ts
+++ b/src/data/repositories/LandingNodeDefaultRepository.ts
@@ -32,7 +32,12 @@ export class LandingNodeDefaultRepository implements LandingNodeRepository {
                 : [];
 
             const validations = roots.map(root =>
-                LandingNodeModel.decode(buildDomainLandingNode(root, _.flatten(persisted)))
+                LandingNodeModel.decode(
+                    buildDomainLandingNode(
+                        root,
+                        persisted.find(model => model.find(item => item.parent === root.id)) ?? []
+                    )
+                )
             );
 
             _.forEach(validations, validation => {


ifoche (approved):
thanks @deeonwuli

src/data/repositories/LandingNodeDefaultRepository.ts
```diff
@@ -32,7 +32,12 @@ export class LandingNodeDefaultRepository implements LandingNodeRepository {
                 : [];
 
             const validations = roots.map(root =>
-                LandingNodeModel.decode(buildDomainLandingNode(root, _.flatten(persisted)))
+                LandingNodeModel.decode(
+                    buildDomainLandingNode(
+                        root,
+                        persisted.find(model => model.find(item => item.parent === root.id)) ?? []
```
p3rcypj, line 38: I see the point of passing only the items on the same array as the root is 👌🏼. That way, only sections owned by the root are built on the root. But it seems the key issue is that, for some reason, there is an identic section repeated in each one of the root's `[]` in the `dataStore/landing-pages`; that's why doing `_.flatten` was making 5 repeated sections when landing on home page app. Another patch would be `_(persisted).flatten().uniqBy(({ id }) => id).value()`, because the problem is the duplicates.

So here there are two key points on the real problem: we shouldn't have duplicates in the `dataStore`, and we shouldn't have a section/sub-section/category in another's root array that doesn't belong to it.

@ifoche suggested that this could be happening due to imports/cloning landing pages. Indeed, I did try in another env "create landing -> create section -> export landing -> import landing"; and it was populating the created section to each landing on the `dataStore` with the same `id`. I don't know what approach would be the best. Maybe @xurxodev can give a hint about "where the validation should be happening"; if it should happen not only when importing landings, but also when retrieving the nodes (maybe on the settings page) that there is an inconsistency on the `dataStore` that should be fixed.

p3rcypj (commented):
Good job @deeonwuli on the solution so the redirect can happen!

It seems that the real problem could be something happening with the import of landings. But the datastore already has inconsistencies, so we really needed something to be able to redirect while we get rid of that.

p3rcypj:
The real issue is having sections on other landings scope and/or duplicated. To solve this we should transform `LandingNode` into a class and perform a validation on the children to verify that every child the parent is indeed the landing itself, and there aren't any duplicates. If so, it should throw an `Error`. By having validation on the entity class, we solve three issues: retrieve, update, and import.

As throwing an error doesn't change the already broken dataStore, we should perform a task to fix the dataStore. We could do this by migrating the data as we do on other projects such as MetadataSync. @ifoche @deeonwuli

PR #29, opened 2024-02-22
feat: Add "internal page" action

p3rcypj:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/8693u555f

### :memo: Implementation

Tasks:

- [x] Added flag "Execute on init" to landing pages
- [x] Added action type "Landing page" to be able to go to another landing page

Validations on update and import:
- Action with type "Landing page" cannot be created if the landing page id doesn't exist
- Action with type "Landing page" cannot be added to landing page that references itself on action

### :video_camera: Screenshots/Screen capture

image
image
image

PR #32, opened 2024-04-23
feat: Clean Arch

p3rcypj:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/8693u555f

### :memo: Implementation

- Removed Dumebi's filter and made the error we had previously on the dataStore no happen again
- Added some comments about clean arch for the next iterations
- Validations for #29 for both landing pages and actions
- Test for landing pages

Validations made:
- Actions (applies to create/edit/import)
  - Action type "page" cannot reference a landing that contains the action itself
  - Action type "page" cannot reference a landing that doesn't exist
- Landing pages (applies to edit/import)`
  - Cannot exist two duplicated nodes (with the same ID)
  - Children nodes must be in their parent tree (landing tree)

The issue we previously experienced was a chain of if / else that was doing several responsibilities, where in one statement it was doing the opposite we wanted, resulting in adding the children node to every landing tree it exists on the dataStore.

### :fire: Testing

Developed with `docker.eyeseetea.com/eyeseetea/dhis2-data:2.37.9-play`

Invalid actions: [invalid-actions.zip](https://github.com/EyeSeeTea/home-page-app-dev/files/15074669/invalid-actions.zip)

Templates to import:
[actions.zip](https://github.com/EyeSeeTea/home-page-app-dev/files/15074739/actions.zip)
[dhis2-landing-page.zip](https://github.com/EyeSeeTea/home-page-app-dev/files/15074740/dhis2-landing-page.zip)
[general-landing-page.zip](https://github.com/EyeSeeTea/home-page-app-dev/files/15074741/general-landing-page.zip)


xurxodev (approved):
thanks @p3rcypj

PR #34, opened 2024-05-06
Add migrations

p3rcypj:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/86943v5z0, https://app.clickup.com/t/8693u555f, https://app.clickup.com/t/8693u9be0

### :memo: Implementation

- Review if duplicate bug to import id fixed without filtering
- Remove the code that was creating the bug
- Validate duplication logic in entity
- Create test to verify the duplication node logic validation
- Verify new functionality about action of type landing
- Implement Migrations (almost ready to be exported as a library)

image

### :fire: Testing

Tested vs `https://dev.eyeseetea.com/play/`


xurxodev (changes requested):
thanks @p3rcypj 

Congratulations good work!

I've tested and it works for me.

I've only minor things to review in the potential folder to be extracted in a lib: d2-migrations.

If we are going to use Clean Architecture within of this folder I would rather some changes:
- Create our typical structure
   - data
      - DataStoreStorage.ts
   - domain
      - repositories
         - MigrationStorage
      - entities
         - Migration.ts
      - use-cases
        - MigrationsRunner -> RunMigrations (and review what methods are necessary as public)
   - presentation
      - Migrations.tsx
      - useMigrations.ts
   - types
     - d2-api.ts
     - utils.ts
   - index.ts
   - utils.ts

xurxodev (approved):
thanks @p3rcypj

PR #35, opened 2024-05-22
Feat/homepage app improvements

deeonwuli:
### :pushpin: References

**Issue:** Closes
- https://app.clickup.com/t/8694gpmmj
- https://app.clickup.com/t/8694f1ern
- https://app.clickup.com/t/8694f1ep7
- https://app.clickup.com/t/8694mkn1x

Screenshot 2024-05-23 at 11 13 51

### :fire: Testing
`REACT_APP_DHIS2_BASE_URL=https://dev.eyeseetea.com/who-dev-238/`
`docker.eyeseetea.com/samaritans/dhis2-data:2.36.11.1-sp-ip-test`

#8694gpmmj
#8694f1ern
#8694f1ep7
#8694mkn1x

### DIFF
--- a/src/domain/entities/Action.ts
+++ b/src/domain/entities/Action.ts
@@ -118,7 +118,7 @@ export const getPageActions = (
 
                 const hasUserAccess = actionUsers.includes(user.id);
                 const hasUserGroupAccess = _.intersection(actionUserGroups, userGroupIds).length > 0;
-                const hasPublicAccess = action.publicAccess && action.publicAccess !== "--------";
+                const hasPublicAccess = Boolean(action.publicAccess) && action.publicAccess !== "--------";
 
                 return hasUserAccess || hasUserGroupAccess || hasPublicAccess;
             })
--- a/src/domain/entities/Action.ts
+++ b/src/domain/entities/Action.ts
@@ -118,7 +118,7 @@ export const getPageActions = (
 
                 const hasUserAccess = actionUsers.includes(user.id);
                 const hasUserGroupAccess = _.intersection(actionUserGroups, userGroupIds).length > 0;
-                const hasPublicAccess = action.publicAccess !== "--------";
+                const hasPublicAccess = action.publicAccess && action.publicAccess !== "--------";
 
                 return hasUserAccess || hasUserGroupAccess || hasPublicAccess;
             })
--- a/src/domain/entities/LandingNode.ts
+++ b/src/domain/entities/LandingNode.ts
@@ -103,17 +103,17 @@ export const updateLandingNodes = (
 const applyFavicon = (parent: LandingNode): LandingNode => {
     const spreadFaviconToChildren = (children: LandingNode[], favicon: string): LandingNode[] => {
         return _.map(children, child => {
-            const updatedChild: LandingNode = { ...child, favicon };
-            if (child.children) {
-                updatedChild.children = spreadFaviconToChildren(child.children, favicon);
-            }
-            return updatedChild;
+            return {
+                ...child,
+                favicon: favicon,
+                children: spreadFaviconToChildren(child.children, favicon),
+            };
         });
     };
 
     return {
         ...parent,
-        children: parent.children ? spreadFaviconToChildren(parent.children, parent.favicon) : [],
+        children: spreadFaviconToChildren(parent.children, parent.favicon),
     };
 };
 
--- a/src/domain/entities/LandingNode.ts
+++ b/src/domain/entities/LandingNode.ts
@@ -71,11 +71,7 @@ export const buildOrderedLandingNodes = (nodes: LandingNode[]): OrderedLandingNo
     }));
 };
 
-export const updateLandingNodes = (
-    nodes: LandingNode[],
-    permissions: LandingPagePermission[],
-    user: User
-): LandingNode[] => {
+function updateLandingNodes(nodes: LandingNode[], permissions: LandingPagePermission[], user: User): LandingNode[] {
     const updatedNodes = _(nodes)
         .map(node => {
             const pagePermission = permissions?.find(permission => permission.id === node.id);
@@ -98,25 +94,24 @@ export const updateLandingNodes = (
         .compact()
         .value();
 
-    return updatedNodes.map(node => applyFavicon(node));
-};
+    return updatedNodes;
+}
 
-const applyFavicon = (parent: LandingNode): LandingNode => {
-    const spreadFaviconToChildren = (children: LandingNode[], favicon: string): LandingNode[] => {
-        return _.map(children, child => {
-            return {
-                ...child,
-                favicon: favicon,
-                children: spreadFaviconToChildren(child.children, favicon),
-            };
-        });
-    };
+function applyFavicon(children: LandingNode[], favicon: string): LandingNode[] {
+    return _.map(children, child => {
+        return {
+            ...child,
+            favicon: favicon,
+            children: applyFavicon(child.children, favicon),
+        };
+    });
+}
 
-    return {
-        ...parent,
-        children: spreadFaviconToChildren(parent.children, parent.favicon),
-    };
-};
+export function updateLandings(nodes: LandingNode[], permissions: LandingPagePermission[], user: User): LandingNode[] {
+    const landings = updateLandingNodes(nodes, permissions, user);
+
+    return landings.map(landing => ({ ...landing, children: applyFavicon(landing.children, landing.favicon) }));
+}
 
 // Return
 // a redirect URL if there is only one visible action on primary nodes
--- a/src/domain/entities/LandingNode.ts
+++ b/src/domain/entities/LandingNode.ts
@@ -24,6 +24,7 @@ export interface LandingNode {
     type: LandingNodeType;
     icon: string;
     iconLocation: string;
+    favicon: string;
     pageRendering: LandingNodePageRendering | undefined;
     order: number | undefined;
     name: TranslatableText;
@@ -42,6 +43,7 @@ export const LandingNodeModel: Codec<LandingNode> = Schema.object({
     type: LandingPageNodeTypeModel,
     icon: Schema.optionalSafe(Schema.string, ""),
     iconLocation: Schema.optionalSafe(Schema.string, ""),
+    favicon: Schema.optionalSafe(Schema.string, ""),
     pageRendering: Schema.optional(LandingPageNodePageRenderingModel),
     order: Schema.optional(Schema.integer),
     name: TranslatableTextModel,
--- a/src/domain/entities/LandingNode.ts
+++ b/src/domain/entities/LandingNode.ts
@@ -97,12 +97,12 @@ function updateLandingNodes(nodes: LandingNode[], permissions: LandingPagePermis
     return updatedNodes;
 }
 
-function applyFavicon(children: LandingNode[], favicon: string): LandingNode[] {
+function spreadFavicon(children: LandingNode[], favicon: string): LandingNode[] {
     return _.map(children, child => {
         return {
             ...child,
             favicon: favicon,
-            children: applyFavicon(child.children, favicon),
+            children: spreadFavicon(child.children, favicon),
         };
     });
 }
@@ -110,7 +110,7 @@ function applyFavicon(children: LandingNode[], favicon: string): LandingNode[] {
 export function updateLandings(nodes: LandingNode[], permissions: LandingPagePermission[], user: User): LandingNode[] {
     const landings = updateLandingNodes(nodes, permissions, user);
 
-    return landings.map(landing => ({ ...landing, children: applyFavicon(landing.children, landing.favicon) }));
+    return landings.map(landing => ({ ...landing, children: spreadFavicon(landing.children, landing.favicon) }));
 }
 
 // Return
--- a/src/domain/entities/LandingNode.ts
+++ b/src/domain/entities/LandingNode.ts
@@ -75,7 +75,7 @@ export const updateLandingNodes = (
     permissions: LandingPagePermission[],
     user: User
 ): LandingNode[] => {
-    return _(nodes)
+    const updatedNodes = _(nodes)
         .map(node => {
             const pagePermission = permissions?.find(permission => permission.id === node.id);
 
@@ -96,6 +96,25 @@ export const updateLandingNodes = (
         })
         .compact()
         .value();
+
+    return updatedNodes.map(node => applyFavicon(node));
+};
+
+const applyFavicon = (parent: LandingNode): LandingNode => {
+    const spreadFaviconToChildren = (children: LandingNode[], favicon: string): LandingNode[] => {
+        return _.map(children, child => {
+            const updatedChild: LandingNode = { ...child, favicon };
+            if (child.children) {
+                updatedChild.children = spreadFaviconToChildren(child.children, favicon);
+            }
+            return updatedChild;
+        });
+    };
+
+    return {
+        ...parent,
+        children: parent.children ? spreadFaviconToChildren(parent.children, parent.favicon) : [],
+    };
 };
 
 /* Return a redirect URL if there is only one visible action on primary nodes */
--- a/src/domain/entities/LandingNode.ts
+++ b/src/domain/entities/LandingNode.ts
@@ -4,6 +4,7 @@ import { TranslatableText, TranslatableTextModel } from "./TranslatableText";
 import { LandingPagePermission } from "./Permission";
 import { User } from "./User";
 import { Action, getPageActions } from "./Action";
+import { Maybe } from "../../types/utils";
 
 export const LandingPageNodeTypeModel = Schema.oneOf([
     Schema.exact("root"),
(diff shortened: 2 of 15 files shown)


ifoche:
Task linked: [CU-8694f1ep7 Add Icon Size settings](https://app.clickup.com/t/8694f1ep7)
Task linked: [CU-8694f1ern Add favicon image on settings](https://app.clickup.com/t/8694f1ern)
Task linked: [CU-8694gpmmj Bug: when I click on a sub action and then go back, I go to the main dashboard page, not the landing page I had just came from](https://app.clickup.com/t/8694gpmmj)

p3rcypj:
Task linked: https://app.clickup.com/t/8694mkn1x

p3rcypj:
Thanks @deeonwuli for your work!! 🙌🏻 Just some key points that I merely tested for the favicon and the icon sizes features:

- This size feature shouldn't apply also for action icons?
- Subsections lose the favicon of their parent, I assume we would like to keep showing the favicon of the root's `landingNode`

   image
   image

- The icon height is not being applied on "Available Landing Pages", not sure if we want the feature here also

   image

src/domain/entities/Action.ts
```diff
@@ -118,7 +118,7 @@ export const getPageActions = (
 
                 const hasUserAccess = actionUsers.includes(user.id);
                 const hasUserGroupAccess = _.intersection(actionUserGroups, userGroupIds).length > 0;
-                const hasPublicAccess = action.publicAccess !== "--------";
+                const hasPublicAccess = action.publicAccess && action.publicAccess !== "--------";
```
p3rcypj: [No need to change] It could be `Boolean(action.publicAccess)` to make it more readable, which is because `action.publicAccess` can be `undefined`, otherwise it seems a little strange that both conditions are together. I have done it this way many times so never mind.

src/domain/entities/LandingNode.ts
```diff
@@ -92,6 +96,25 @@ export const updateLandingNodes = (
...
+
+    return updatedNodes.map(node => applyFavicon(node));
+};
+
+const applyFavicon = (parent: LandingNode): LandingNode => {
+    const spreadFaviconToChildren = (children: LandingNode[], favicon: string): LandingNode[] => {
+        return _.map(children, child => {
+            const updatedChild: LandingNode = { ...child, favicon };
+            if (child.children) {
+                updatedChild.children = spreadFaviconToChildren(child.children, favicon);
```
p3rcypj: Assign the children prop on constant declaration

src/webapp/components/landing-page-edit-dialog/LandingPageEditDialog.tsx
```diff
@@ -111,11 +114,40 @@ export const LandingPageEditDialog: React.FC<LandingPageEditDialogProps> = props
     };
 
     const handleFileUpload = useCallback(
```
p3rcypj: I would rename it to `handleImageFileUpload` because we are using image properties within

src/webapp/components/landing-page-edit-dialog/LandingPageEditDialog.tsx
```diff
@@ -111,11 +114,40 @@ export const LandingPageEditDialog: React.FC<LandingPageEditDialogProps> = props
...
+                    img.onload = () => {
+                        const width = img.width;
+                        const height = img.height;
+                        const aspectRatio = width / height;
+
+                        setWarnings([]);
+                        const newWarnings: string[] = [];
+
+                        if (fileType === "favicon") {
+                            if (aspectRatio !== FAVICON_ASPECT_RATIO) {
```
p3rcypj: Although it is unfortunate, we have some client dashboards where they use icons with a one-pixel difference like 400x401 🥴. We could leave a margin of error of a maximum of 5 pixels or something like that. For example an error margin of `~0.015`

adrianq (changes requested):
@deeonwuli Thanks for the work! As mentioned by Slack, when only one internal page is available/accessible (and no action) we should send the user straight to the internal page. The scenario is reproducible in the docker I shared with you

This bug https://app.clickup.com/t/8694gpmmj is still happening

src/webapp/components/landing-page-edit-dialog/LandingPageEditDialog.tsx
```diff
@@ -111,11 +114,40 @@ export const LandingPageEditDialog: React.FC<LandingPageEditDialogProps> = props
...
+                                newWarnings.push("Please ensure that your favicon has a 1:1 aspect ratio.");
```
p3rcypj: Use interpolation if you declared `FAVICON_ASPECT_RATIO`. But as we have the result of the division as value, create both the fraction and the result/value to use it in the interpolation and on the if statement

src/webapp/components/landing-page-edit-dialog/LandingPageEditDialog.tsx
```diff
@@ -111,11 +114,40 @@ export const LandingPageEditDialog: React.FC<LandingPageEditDialogProps> = props
...
+                            }
+
+                            if (width > FAVICON_MAX_SIZE || height > FAVICON_MAX_SIZE) {
+                                newWarnings.push("Please use an icon of 128x128 pixels or smaller.");
```
p3rcypj: Interpolation

src/webapp/components/landing-page-edit-dialog/LandingPageEditDialog.tsx
```diff
@@ -278,6 +351,9 @@ export const LandingPageEditDialog: React.FC<LandingPageEditDialogProps> = props
     );
 };
 
+const FAVICON_ASPECT_RATIO = 1;
+const FAVICON_MAX_SIZE = 128;
```
p3rcypj: I would increase the max size of the favicon, as normally we even use 512x512 for favicons frequently in web environment ([MDN: Define your app icons](https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/How_to/Define_app_icons))

src/domain/entities/LandingNode.ts
```diff
@@ -92,6 +96,25 @@ export const updateLandingNodes = (
...
+            }
+            return updatedChild;
+        });
+    };
+
+    return {
+        ...parent,
+        children: parent.children ? spreadFaviconToChildren(parent.children, parent.favicon) : [],
```
p3rcypj:
```
Boolean([]) // true
```

No need to have this ternary operator (it even made me doubt if in runtime it could be an undefined, but checked [buildLandingNode](https://github.com/EyeSeeTea/home-page-app-dev/blob/a82d9b60948882e38a8ae025b126f813919d3717/src/data/repositories/LandingNodeDefaultRepository.ts#L126))

src/webapp/components/landing-page-edit-dialog/LandingPageEditDialog.tsx
deeonwuli: when i increase the size of the favicon, i can't view it in the browser and it defaults to the current page's icon

src/webapp/components/landing-page-edit-dialog/LandingPageEditDialog.tsx
```diff
@@ -111,11 +114,40 @@ export const LandingPageEditDialog: React.FC<LandingPageEditDialogProps> = props
...
+                const reader = new FileReader();
+                reader.onload = e => {
+                    const img = new Image();
+                    img.onload = () => {
+                        const width = img.width;
+                        const height = img.height;
+                        const aspectRatio = width / height;
+
+                        setWarnings([]);
+                        const newWarnings: string[] = [];
```
p3rcypj: Use functional programming instead of declaring `[]` and later using `.push`

src/webapp/components/landing-page-edit-dialog/LandingPageEditDialog.tsx
```diff
@@ -111,11 +114,40 @@ export const LandingPageEditDialog: React.FC<LandingPageEditDialogProps> = props
...
+                        if (fileType === "favicon") {
+                            if (aspectRatio !== FAVICON_ASPECT_RATIO) {
+                                newWarnings.push("Please ensure that your favicon has a 1:1 aspect ratio.");
+                            }
+
+                            if (width > FAVICON_MAX_SIZE || height > FAVICON_MAX_SIZE) {
+                                newWarnings.push("Please use an icon of 128x128 pixels or smaller.");
+                            }
+
+                            newWarnings.length !== 0 && setWarnings(newWarnings);
```
p3rcypj: Using functional programming to assign directly the warnings you will no need to make this check later.

P.S.: `!_.isEmpty(newWarnings)` with lodash

src/webapp/components/landing-page-edit-dialog/LandingPageEditDialog.tsx
```diff
@@ -111,11 +114,40 @@ export const LandingPageEditDialog: React.FC<LandingPageEditDialogProps> = props
...
     const handleFileUpload = useCallback(
-        (event: ChangeEvent<HTMLInputElement>) => {
+        (event: ChangeEvent<HTMLInputElement>, fileType: keyof LandingNode) => {
             const file = event.target.files ? event.target.files[0] : undefined;
+
             file?.arrayBuffer().then(async data => {
+                const reader = new FileReader();
+                reader.onload = e => {
+                    const img = new Image();
+                    img.onload = () => {
```
p3rcypj: Extract the function to a constant and give it a descriptive name for it

src/webapp/components/landing-page-edit-dialog/LandingPageEditDialog.tsx
```diff
@@ -186,7 +218,7 @@ export const LandingPageEditDialog: React.FC<LandingPageEditDialogProps> = props
                         </IconContainer>
                     ) : null}
 
-                    <FileInput type="file" onChange={handleFileUpload} />
+                    <FileInput type="file" onChange={event => handleFileUpload(event, "icon")} />
```
p3rcypj: `useCallback`

src/webapp/components/landing-page-edit-dialog/LandingPageEditDialog.tsx
```diff
@@ -202,8 +234,49 @@ export const LandingPageEditDialog: React.FC<LandingPageEditDialogProps> = props
...
+                <OptionContainer>
+                    {["small", "medium", "large"].map((size, i) => (
+                        <Button
+                            key={i}
+                            color={
+                                size === value.iconSize || (!value.iconSize && size === "small") ? "primary" : "default"
+                            }
+                            variant="contained"
+                            value={size}
+                            onClick={() => setValue(value => ({ ...value, iconSize: size }))}
```
p3rcypj: `useCallback`

src/webapp/components/landing-page-edit-dialog/LandingPageEditDialog.tsx
```diff
@@ -202,8 +234,49 @@ export const LandingPageEditDialog: React.FC<LandingPageEditDialogProps> = props
...
+                    <h3>{i18n.t("Favicon")}</h3>
+
+                    <IconUpload>
+                        {value.favicon ? (
+                            <IconContainer>
+                                <img src={value.favicon} alt={`Page favicon`} />
+                            </IconContainer>
+                        ) : null}
+
+                        <FileInput type="file" onChange={event => handleFileUpload(event, "favicon")} />
```
p3rcypj: `useCallback`

src/webapp/components/landing-page-edit-dialog/LandingPageEditDialog.tsx
```diff
@@ -202,8 +234,49 @@ export const LandingPageEditDialog: React.FC<LandingPageEditDialogProps> = props
...
+                    </IconUpload>
+
+                    {warnings.length > 0 && (
```
p3rcypj: Use `!_.isEmpty()`

src/webapp/components/landing-page-edit-dialog/useImageFileUpload.tsx
```diff
@@ -0,0 +1,93 @@
+import { ChangeEvent, useCallback, useState } from "react";
+import { useAppContext } from "../../contexts/app-context";
+import { LandingNode } from "../../../domain/entities/LandingNode";
+import i18n from "../../../locales";
+
+export default function useImageFileUpload(updateNode: (value: React.SetStateAction<LandingNode>) => void) {
```
p3rcypj, line 7: Good move! 🙌🏼

src/webapp/components/landing-page-edit-dialog/LandingPageEditDialog.tsx
```diff
@@ -202,8 +200,49 @@ export const LandingPageEditDialog: React.FC<LandingPageEditDialogProps> = props
...
+                <OptionContainer>
+                    {["small", "medium", "large"].map((size, i) => (
+                        <Button
+                            key={i}
+                            color={
+                                size === value.iconSize || (!value.iconSize && size === "small") ? "primary" : "default"
+                            }
+                            variant="contained"
+                            value={size}
+                            onClick={() => onChangeIconSize(size)}
```
p3rcypj, line 216: **[No need to change]** In this case doesn't apply because these three buttons will re-render anyway because of `[].map()`. But in other case, these three buttons would also re-render each time because of the creation of the arrow function.

```javascript
(()=>{}) === (()=>{}) // output: false
```

React will treat the props as different and would cause the button to re-render each time. And same applies for arrays and objects:

```javascript
[] === [] // false
({}) == ({}) // false
Object() === Object() // false
```

A more meaningful example would be:

```javascript
["small", "medium", "large"] === ["small", "medium", "large"] // output: false
```

That's why these buttons will be re-rendered each time. A common solution would be to declare the array outside of the component scope (as you know we usually do it at the end of the file along other constants). And also adding the `() =>` part to the `useCallback`.

This was a heavy headache for me to understand... and it took me 2 years to realize 🫠. This doesn't apply to conditions where the value is a string or a number because: `"hello" === "hello" // true`. That's why is not necessary for the color prop. It also took me a lot of reviews from Arnau and for sure if nobody reviews my code I will not do it perfectly on each given prop :')

Another thing to consider here is why I put **[No need to change]**; this button from Material UI is straightforward and I'm sure it will not cost much effort to re-render. But this is really a MUST if the component is a custom one with complex functionality, for example, this component itself is a more complex one `LandingPageEditDialog` than a button. So don't change it but be aware of it 🥷🏼.

src/webapp/components/landing-page-edit-dialog/useImageFileUpload.tsx
```diff
@@ -0,0 +1,93 @@
...
+                            ? [
+                                  i18n.t("Please ensure that your favicon has a {{aspectRatio}} aspect ratio.", {
+                                      aspectRatio: FAVICON_ASPECT_RATIO.fraction,
+                                  }),
+                              ]
+                            : []),
+                    ];
+
+                    const imageDimensionWarnings = [
+                        ...(!isWithinMargin(width) || !isWithinMargin(height)
```
p3rcypj: `!(isWithinMargin(width) && isWithinMargin(height)` would be more readable

src/webapp/components/landing-page-edit-dialog/useImageFileUpload.tsx
```diff
@@ -0,0 +1,93 @@
...
+        reader.onload = e => {
+            const img = new Image();
+            img.onload = () => {
+                const width = img.width;
+                const height = img.height;
+                const aspectRatio = width / height;
+
+                setFaviconWarnings([]);
+                if (fileType === "favicon") {
+                    const aspectRatioWarnings = [
```
p3rcypj, line 23: Good work! **[No need to change]** There is a lot of ways we could declare this, another way a bit more readable would be for example (**that I didn't test it**, it should work but I didn't check):

```typescript
const warnings = _.compact([
    !isAspectRatioWithinMargin(aspectRatio) &&
        i18n.t("Please ensure that your favicon has a {{aspectRatio}} aspect ratio.", {
            aspectRatio: FAVICON_ASPECT_RATIO.fraction,
        }),
    !(isWithinMargin(width) && isWithinMargin(height)) &&
        i18n.t("Please use an icon of {{max}}x{{max}} pixels or smaller.", {
            max: FAVICON_MAX_SIZE,
        }),
]);

setFaviconWarnings(warnings);
```

src/webapp/components/landing-page-edit-dialog/useImageFileUpload.tsx
```diff
@@ -0,0 +1,93 @@
...
+            });
+        },
+        [compositionRoot.instance, updateNode]
+    );
+
+    const uploadIcon = useCallback(
+        (event: ChangeEvent<HTMLInputElement>) => handleImageFileUpload(event, "icon"),
+        [handleImageFileUpload]
+    );
+    const uploadFavicon = useCallback(
```
p3rcypj, line 72: Add a blank line here and there to make it a bit more faster to read

src/webapp/components/landing-page-edit-dialog/useImageFileUpload.tsx
```diff
@@ -0,0 +1,93 @@
...
+const FAVICON_ASPECT_RATIO = { fraction: "1:1", value: 1 };
+const FAVICON_MAX_SIZE = 128;
+
+const ALLOWED_MARGIN = 5;
+const ASPECT_RATIO_MARGIN = 0.015;
+
+const isWithinMargin = (dimension: number): boolean => {
+    return dimension >= FAVICON_MAX_SIZE - ALLOWED_MARGIN && dimension <= FAVICON_MAX_SIZE + ALLOWED_MARGIN;
+};
+const isAspectRatioWithinMargin = (aspectRatio: number): boolean =>
```
p3rcypj, line 94: Blank line

src/webapp/components/landing-page-edit-dialog/useImageFileUpload.tsx
```diff
@@ -0,0 +1,93 @@
...
+            if (e.target?.result) {
+                img.src = e.target.result as string;
+            }
+        };
+        reader.readAsDataURL(file);
+    };
+
+    const handleImageFileUpload = useCallback(
+        (event: ChangeEvent<HTMLInputElement>, fileType: keyof LandingNode) => {
+            const file = event.target.files ? event.target.files[0] : undefined;
```
p3rcypj: With `_.first()` you will not need ternary operator ^^

p3rcypj:
`LandingPageEditDialog.tsx` is increasing, I added two comments TODO for the next person to edit this file

src/domain/entities/LandingNode.ts
```diff
@@ -92,44 +97,69 @@ export const updateLandingNodes = (
...
+export function getPrimaryRedirectNodes(
     landingNode: LandingNode,
     options: { actions: Action[]; user: User }
-): Url | undefined {
+): { redirectUrl: Maybe<Url>; redirectPageId: Maybe<string> } {
     const { actions, user } = options;
 
     const actionsById = _.keyBy(actions, action => action.id);
     const showAllActions = false;
     const isRoot = true;
```
p3rcypj, line 129: **[Don't change it]** Good for describing what param is👌🏼 When the number of params increases as in this case, we use `(options: { })` when declaring the function

src/domain/entities/LandingNode.ts
```diff
@@ -92,44 +97,69 @@ export const updateLandingNodes = (
...
+                ...child,
+                favicon: favicon,
+                children: spreadFaviconToChildren(child.children, favicon),
+            };
+        });
+    };
+
+    return {
+        ...parent,
+        children: spreadFaviconToChildren(parent.children, parent.favicon),
```
p3rcypj: **[Already changed]** Do you see the pattern? indeed the `updatedNodes` where we want to spread the favicon are only the landings themselves. So we can put the functionality of `applyFavicon` inside the `.map` return and simplify the recursive function. Also, as we are executing the recursive function inside another recursive function `updateLandingNodes` updating the children N children times. The best option would be to traverse the tree once applying the favicon simultaneously. But since we are running out of time we can also traverse it twice. Check `updateLandings` L110

p3rcypj (approved):
Just some very quick comments, the other ones are just informative. By the way, I was not able to be redirected to CMAM landing / neither data entry app, what behaviour should be expected?

src/webapp/components/landing-page-edit-dialog/LandingPageEditDialog.tsx
p3rcypj: I've checked in Chrome, Edge, and Firefox, and it works even with 1024x1024... we should ask a third person @adrianq  to check if user is able to see favicons bigger than 128

src/webapp/components/landing-page-edit-dialog/LandingPageEditDialog.tsx
deeonwuli: got it. thanks @p3rcypj 😀

PR #36, opened 2024-06-04
set app namespace in appcontext

eperedo:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/8694qwkw6

### :memo: Implementation

- [x] Update i18n types
- [x] set default namespace in `app-context.ts`

### The Problem

In our applications we have a `locales/index.js` file that set the namespace for the translations.

```js
const namespace = 'homepage-app';
...
i18n.setDefaultNamespace(namespace);
```

The problem is that we use libraries like `d2-ui-components` or `feedback-component` that also have their own `locale/index.js` files:

```js
const namespace = 'd2-ui-components';
...
i18n.setDefaultNamespace(namespace);
```

so eventually the application namespace will be overwritten by one of these libraries.

### Fast solution

Force the whole application to use the namespace of the app by doing it in the `app-context.ts` file

```ts
i18n.setDefaultNamespace("homepage-app");
```

if you get the error: `Property 'setDefaultNamespace' does not exist on type 'typeof import("@dhis2/d2-i18n")'`
look for the `i18n.d.ts` file and add the type in there:

```ts
declare module "@dhis2/d2-i18n" {
    // ...
    export function setDefaultNamespace(namespace: string);
}
```

### :fire: Testing

For testing purposes I'm only adding a translation for the string `Available Home Pages`
image

#8694qwkw6

### DIFF
--- a/i18n/en.pot
+++ b/i18n/en.pot
@@ -5,8 +5,8 @@ msgstr ""
 "Content-Type: text/plain; charset=utf-8\n"
 "Content-Transfer-Encoding: 8bit\n"
 "Plural-Forms: nplurals=2; plural=(n != 1)\n"
-"POT-Creation-Date: 2024-05-06T08:44:12.108Z\n"
-"PO-Revision-Date: 2024-05-06T08:44:12.108Z\n"
+"POT-Creation-Date: 2024-06-04T15:47:01.351Z\n"
+"PO-Revision-Date: 2024-06-04T15:47:01.351Z\n"
 
 msgid "There are pending migrations"
 msgstr ""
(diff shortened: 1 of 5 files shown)


ifoche:
Task linked: [CU-8694qwkw6 Fix translations conflict between home page and feedback component](https://app.clickup.com/t/8694qwkw6)

tokland (approved):
All good!

tokland:
Related: https://github.com/EyeSeeTea/dhis2-app-skeleton/pull/91/commits/54fcc221503d23f3bf4490d4315679b4cf60034b

PR #39, opened 2025-01-14
[fix] LandingNode delete error

deeonwuli:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/86979ke34

### :memo: Implementation
- [x] `deleteNodes` method now handles empty models 

https://github.com/user-attachments/assets/1afc450a-d536-446e-b74e-d51e3ef01d7b

### :fire: Testing
`REACT_APP_DHIS2_BASE_URL=https://dev.eyeseetea.com/who-dev-238`

#86979ke34

### DIFF
--- a/src/data/repositories/LandingNodeDefaultRepository.ts
+++ b/src/data/repositories/LandingNodeDefaultRepository.ts
@@ -98,17 +98,22 @@ export class LandingNodeDefaultRepository implements LandingNodeRepository {
     public async deleteNodes(ids: string[]): Promise<void> {
         const nodes = (await this.storageClient.getObject<PersistedLandingPage[]>(Namespaces.LANDING_PAGES)) ?? [];
 
-        const newNodes = nodes.map(models => {
-            const root = models.find(model => model.type === "root");
-            if (!root) throw new Error("No value for root");
+        const newNodes = _(nodes)
+            .map(models => {
+                if (_.isEmpty(models)) return undefined;
 
-            const node = LandingNodeModel.decode(buildLandingNode(root, models)).toMaybe().extract();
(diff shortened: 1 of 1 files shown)


ifoche:
Task linked: [CU-86979ke34 Home page fails upon deleting sections or subsections](https://app.clickup.com/t/86979ke34)

eperedo (approved):
Code looks good, but I was able to delete landing pages/sections from the `development` branch so probably I'm missing something. Not really sure how we can hit the `"No value for root"` error.

deeonwuli:
> Code looks good, but I was able to delete landing pages/sections from the `development` branch so probably I'm missing something. Not really sure how we can hit the `"No value for root"` error.

For whatever reason, there was an empty array in the datastore that was causing the bug. This fix has removed the empty array that would result in the `"No value for root"` error.

eperedo:
> > Code looks good, but I was able to delete landing pages/sections from the `development` branch so probably I'm missing something. Not really sure how we can hit the `"No value for root"` error.
> 
> For whatever reason, there was an empty array in the datastore that was causing the bug. This fix has removed the empty array that would result in the `"No value for root"` error.

Ohh ok thanks! Totally outside of the scope of this PR, but it would be good to keep an eye on how this could happen maybe we have a bug in other part of the app. @MiquelAdell

MiquelAdell:
So @deeonwuli the change in the code makes it not fail if we have the empty array?

I'll create a new task for the empty array to keep a reference for this but its one of those that may be difficult to reproduce

Thanks!

eperedo:
> So @deeonwuli the change in the code makes it not fail if we have the empty array?

Yes, this PR fixed the problem.

PR #42, opened 2025-04-09
Notifications

gqcorneby:
### :pushpin: References 

-   **Issue:** Closes #8698n31xh [[Homepage App] Implement notifications system](https://app.clickup.com/t/8698n31xh)

### :memo: Implementation (Fetching and displaying. Mark as read)
- updated `DataStoreStorageClient` to accept a namespace.  use `dataStoreNamespace` by default.
- Notification domain and data layer
- fetch, show, and updates notifications
- update some reference to i18n

https://github.com/user-attachments/assets/51a6050e-4a88-4410-8c49-b7932785d08c

wildcard set as ANDROID, content not displayed
image


ifoche:
Task linked: [CU-8698n31xh [Homepage App] Implement notifications system](https://app.clickup.com/t/8698n31xh)

src/domain/repositories/NotificationRepository.ts
```diff
@@ -0,0 +1,15 @@
+import { FutureData } from "../types/Future";
+import { Notification, NotificationWildcardType } from "../entities/Notification";
+import { Maybe } from "../../types/utils";
+import { User } from "../entities/User";
+
+export interface NotificationRepository {
+    list(options: NotificationListOptions): FutureData<Notification[]>;
+    save(notifications: Partial<Notification>[]): FutureData<void>;
```
tokland: [clean arch] the entity should be used unmodified in arguments and return values -> `notifications: Notification>[]`. I guess these will entail some changes elsewhere, but should not use a Partial here.

src/domain/usecases/ListCurrentUserNotificationsUseCase.ts
```diff
@@ -0,0 +1,19 @@
...
+import { Notification, NotificationWildcard } from "../entities/Notification";
+import { UserRepository } from "../repositories/UserRepository";
+
+export class ListCurrentUserNotificationsUseCase implements UseCase {
+    constructor(private notificationRepository: NotificationRepository, private userRepository: UserRepository) {}
+
+    public execute(): FutureData<Notification[]> {
+        return this.userRepository.getCurrentUser().flatMap(user => {
+            return this.notificationRepository.list({
+                user,
```
tokland: very minor, just to keep code aligned, we typically write 'user: user'. That's nice for code navigation, you can directly go to the type (left) or the value (right). Otherwise, you get a list of them mixed.

src/webapp/components/user-notification/UserNotificationDialog.tsx
```diff
@@ -0,0 +1,33 @@
...
+import { Notification } from "../../../domain/entities/Notification";
+import i18n from "../../../utils/i18n";
+import { NotificationContent } from "./NotificationContent";
+
+export const UserNotificationDialog: React.FC<UserNotificationDialogProps> = ({
+    notifications,
+    onClose,
+    onConfirm,
+}) => {
+    const content = notifications.map(({ content }) => content).join("\n\n");
```
tokland: we typically favor writing `props` and de-structuring it in the first line of the component. typically, more readable, and may save some LOCs. However, this is subjective, keep it as you prefer.

src/webapp/pages/App.tsx
```diff
@@ -1,17 +1,61 @@
...
+                        await compositionRoot.notifications.readUserNotifications(notifications).toPromise();
+                        continueLoading();
+                    },
+                });
+            } else {
+                continueLoading();
+            }
+        }
+        setup();
+    }, [baseUrl, locale]);
```
tokland, line 27: let's move this useEffect to a custom hook. It lightens the component, but also makes it more declarative, which is specially important for useEffect blocks, which have no name so we need to read the contents to see what it's doing.

src/domain/usecases/ReadCurrentUserNotifications.ts
```diff
@@ -0,0 +1,22 @@
+import { UseCase } from "./UseCase";
+import { NotificationRepository } from "../repositories/NotificationRepository";
+import { FutureData } from "../types/Future";
+import { User } from "../entities/User";
+import { Notification } from "../entities/Notification";
+import { UserRepository } from "../repositories/UserRepository";
+
+export class ReadCurrentUserNotifications implements UseCase {
```
tokland: Sufix "UseCase" missing.

tokland: For some time, we wrote "implements UseCase," which, in principle, is a good idea. It ensures that you have an "execute" method and makes it declarative. However, practice shows that it's inconvenient. If you go to the "execute" method and perform a "Go to definition" or "Go to references," you will see an enormous list. That's due to the "implements". I'd remove it for new and updated code.

Note: `getExecute` also seemed like a good idea, but it's not, the code navigation is also hindered. Of course no need to change it, but just for the record.

src/domain/entities/Notification.ts
```diff
@@ -0,0 +1,29 @@
...
+    id: string;
+    name: string;
+    date: Date;
+};
+
+export const NotificationWildcard = {
+    ALL: "ALL",
+    ANDROID: "ANDROID",
+    WEB: "WEB",
+} as const;
```
tokland, line 28: I was about to suggest using an "enum" but it's pretty much the same and the community is somewhat divided about its usage (because it generates runtime code).

src/data/clients/storage/DataStoreStorageClient.ts
```diff
@@ -8,11 +8,10 @@ export class DataStoreStorageClient extends StorageClient {
     private api: D2Api;
     private dataStore: DataStore;
 
-    constructor(type: "user" | "global", instance: Instance) {
+    constructor(type: "user" | "global", instance: Instance, namespace: string = dataStoreNamespace) {
```
tokland: when we have options with very plain types (boolean, string), we typically wrap it. In this case we can wrap everything in an options object, it will look more declarative.

src/data/repositories/NotificationDefaultRepository.ts
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    public save(notifications: Partial<Notification>[]): FutureData<void> {
+        return this.list(null)
+            .map(existingNotifs => this.updateNotifications(existingNotifs, notifications))
+            .flatMap(updatedNotifications => this._save(updatedNotifications))
+            .flatMapError(error => {
+                console.error(`Notification (save): ${error}`);
+                return Future.error(i18n.t("An error has occurred while saving notifications"));
+            });
```
tokland: Subjective, but to keep the public methods as terse and declarative as possible, I'd suggest moving the flatMapError to the inner methods (_save, and also _get, it's the same idea).

src/data/repositories/NotificationDefaultRepository.ts
```diff
@@ -0,0 +1,102 @@
...
+            return notifForUser && options?.isRead === isRead;
+        }
+        return notifForUser;
+    }
+
+    private updateNotifications(
+        existingNotifications: Notification[],
+        notificationUpdates: Partial<Notification>[]
+    ): Notification[] {
+        const notificationMap = _.keyBy(notificationUpdates, "id");
```
tokland: Unfortunately, `_.keyBy(notificationUpdates, "id")` is not typ-esafe, because of lodash, we need to use an arrow function `_.keyBy(notificationUpdates, notification => notification.id)`

src/data/repositories/NotificationDefaultRepository.ts
```diff
@@ -0,0 +1,102 @@
...
+                    this.isValidWildcard(notification, options?.wildcard) && this.isForUser(notification, options)
+            )
+            .value();
+    }
+
+    private isValidWildcard(notification: Notification, wildcardOptions: Maybe<NotificationWildcardType[]>): boolean {
+        return (
+            !notification.recipients.wildcard ||
+            !wildcardOptions ||
+            [NotificationWildcard.ALL, ...(wildcardOptions || [])].includes(notification.recipients.wildcard)
```
tokland: Do we need the fallback? TS says `(parameter) wildcardOptions: ("ALL" | "ANDROID" | "WEB")[]`

src/data/repositories/NotificationDefaultRepository.ts
```diff
@@ -0,0 +1,102 @@
...
+    private isForUser(notification: Notification, options?: NotificationListOptions): boolean {
+        if (!options?.user) return true; // If no user filter is provided, return true
+
+        const isForUser = notification.recipients.users.some(({ id }) => id === options.user.id);
+        const isForGroup = notification.recipients.userGroups.some(({ id }) =>
+            options.user.userGroups.some(group => id === group.id)
+        );
+        const notifForUser = isForUser || isForGroup;
+
+        if (typeof options?.isRead === "boolean") {
```
tokland: Typically, having to use `typeof` is a red flag, but I get it, we have `boolean | null | undefined` and it's the simplest way. Striving for declarative code, maybe this?

```typescript
function isBoolean(value: unknown): value is boolean {
  return typeof value === 'boolean';
}
```

tokland (commented):
[code-only review] Minor things:

src/domain/usecases/ReadCurrentUserNotificationsUseCase.ts
```diff
@@ -0,0 +1,41 @@
...
+        const notificationMap = _.keyBy(notifications, notification => notification.id);
+        return this.notificationRepository
+            .list(null)
+            .map(allNotifications => allNotifications.filter(notification => !!notificationMap[notification.id]))
+            .map(notificationsToUpdate =>
+                notificationsToUpdate.map(notification => this.markAsRead(notification, user))
+            );
+    }
+
+    private markAsRead(notification: Notification, user: User): Notification {
```
tokland, line 21: This kind of code, core logic of the entity more than a enterprise logic, can also be in the entity itself.

src/webapp/components/user-notification/UserNotificationDialog.tsx
```diff
@@ -3,11 +3,8 @@ import { Notification } from "../../../domain/entities/Notification";
 import i18n from "../../../utils/i18n";
 import { NotificationContent } from "./NotificationContent";
 
-export const UserNotificationDialog: React.FC<UserNotificationDialogProps> = ({
-    notifications,
-    onClose,
-    onConfirm,
-}) => {
+export const UserNotificationDialog: React.FC<UserNotificationDialogProps> = (props: UserNotificationDialogProps) => {
```
tokland: [no need to change] TS already knows the type of `props` , from `FC<Props>`

src/webapp/hooks/useUserNotifications.ts
```diff
@@ -0,0 +1,49 @@
+import { AppContextProviderProps } from "../contexts/app-context";
+import { useEffect, useState } from "react";
+import { UserNotificationDialogProps } from "../components/user-notification/UserNotificationDialog";
+
+export function useUserNotifications(props: useUserNotificationProps) {
+    const { appContextProps } = props;
+    const [isUserNotifsLoading, setIsUserNotifsLoading] = useState(true);
+    const [userNotificationDialogProps, setUserNotificationDialogProps] = useState<UserNotificationDialogProps | null>(
+        null
+    );
```
tokland: No need to change, but typically initial empty values are handled with "undefined": `useState<UserNotificationDialogProps>()`

tokland (approved):
Just some comments that need no action:

gqcorneby:
Hi @tokland, added some changes to wildcard handling

"ALL", "Android", "Web", "Both",
"ALL" → all users irrespective of groups
"Android" → only android users respecting user groups
"Web", only web users respecting user groups
"Both", both android and web users respecting user groups

I transferred user and isRead filtering to the use case. Only filtering done in the repository now is just the wildcard.

src/domain/usecases/ListCurrentUserNotificationsUseCase.ts
```diff
@@ -1,18 +1,42 @@
...
-        });
+        const notificationFilter: NotificationListOptions = {
+            wildcard: [NotificationWildcard.ALL, NotificationWildcard.WEB, NotificationWildcard.Both],
+        };
+        return this.userRepository
+            .getCurrentUser()
+            .flatMap(user =>
+                this.notificationRepository.list(notificationFilter).map(notifications => ({ user, notifications }))
+            )
+            .map(({ user, notifications }) => this.filterUserNotifications(notifications, user));
```
tokland, line 20: [no need to change] this.filterUserNotifications can be directly called in the flatMap block, but this is also valid.

tokland (approved):

ifoche (changes requested):
@gqcorneby I'm sorry to come back to you so late with this, but I've just tested it functionally and detected some glitch in it. If I have 2 different "ALL" notifications, my user can only see one of them. When I remove one, then the other shows up. The ideal behaviour here would be that you just get as many pop-ups as notifications should reach you, and not only that one of the reaches you. Could you please check it and do the changes to the code accordingly?

Thank you!

ifoche (approved):
thanks @gqcorneby @tokland

PR #43, opened 2025-04-15
add matomo analytics

eperedo:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/8695uy2vv, https://app.clickup.com/t/8698wdywy

### :memo: Implementation

- [x] save matomo url in dataStore
- [x] track page view using matomo js API
- [x] Track analytics via user-triggered events (homepage clicks, landing page clicks, button actions, etc.) instead of react hooks to prevent duplicate event tracking.
- [x] Remove analytics code from domain layer and keep it in the presentation layer.

**Home Page App dashboard**

image

### :fire: Testing

#8695uy2vv
#8698wdywy


ifoche:
Task linked: [CU-8695uy2vv Investigate how to integrate Matomo in the Oslo Android app (OCC fork)](https://app.clickup.com/t/8695uy2vv)

ifoche:
Task linked: [CU-8698wdywy Duplicate pageViews in home page](https://app.clickup.com/t/8698wdywy)

src/webapp/components/inline-input-save/InlineInputSave.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+    return (
+        <SubContainer>
+            <h4>{title}</h4>
+            <GridForm>
+                <TextFieldOnBlur
+                    fullWidth={true}
+                    label={label}
+                    value={value}
+                    onChange={event => onChange(event.target.value)}
```
p3rcypj, line 26: useCallback

src/webapp/components/inline-input-save/InlineInputSave.tsx
```diff
@@ -0,0 +1,35 @@
...
+                    placeholder={placeholder}
+                />
+                <Button onClick={() => onUpdate(value)} color="primary" variant="contained">
```
p3rcypj, line 29: useCallback

src/webapp/components/inline-input-save/InlineInputSave.tsx
p3rcypj: useCallback

src/webapp/components/inline-input-save/InlineInputSave.tsx
p3rcypj: useCallback

src/webapp/pages/App.tsx
```diff
@@ -50,4 +51,24 @@ const Analytics: React.FC = () => {
     return <></>; //return as <script/> seems GA doesn't like that :$
 };
 
+export const MatomoScript = () => {
```
p3rcypj, line 83: This would be hand in hand with the resolution on [#409 (CPR)](https://github.com/EyeSeeTea/central-planning-reporting/pull/409) after the resolution has been made

src/webapp/pages/home/HomePage.tsx
```diff
@@ -263,3 +256,49 @@ function useRedirectOnSinglePrimaryNode(
 
     return { isActive: isActive, currentPage: currentPage };
 }
+
+type UseTrackAnalyticsOnLoadProps = {
+    compositionRoot: CompositionRoot;
+    userLandings: Maybe<LandingNode[]>;
+};
+
+function useTrackAnalyticsOnLoad(props: UseTrackAnalyticsOnLoadProps) {
```
p3rcypj: Extract useHook into a separate file

src/webapp/pages/home/HomePage.tsx
```diff
@@ -263,3 +256,49 @@ function useRedirectOnSinglePrimaryNode(
...
+    React.useEffect(() => {
+        const initLandings = userLandings?.filter(landing => landing.executeOnInit);
+        const pageType = initLandings && initLandings?.length > 1 ? "userLandings" : "singleLanding";
+        if (userLandings && userLandings.length > 1 && pageType === "userLandings") {
+            trackUserLanding(compositionRoot);
+        }
+
+        if (initLandings && initLandings.length > 0 && pageType === "singleLanding") {
+            const cuPage = initLandings[0];
+            if (!cuPage) return;
```
p3rcypj: You could also group the `cuPage` check into the previous `if` statement by assigning the constant before.

src/webapp/pages/home/HomePage.tsx
```diff
@@ -263,3 +256,49 @@ function useRedirectOnSinglePrimaryNode(
...
+
+function trackSingleLanding(landing: LandingNode, compositionRoot: CompositionRoot) {
+    const viewOptions = buildViewOptionsFromLanding(landing);
+    compositionRoot.analytics.sendPageView(viewOptions);
+    compositionRoot.matomo.trackView(viewOptions);
+}
+
+function trackUserLanding(compositionRoot: CompositionRoot) {
+    const viewOptions = {
+        location: `${window.location.hash.split("?")[0]}home-page-app/available-landings`,
```
p3rcypj: `window.location.hash.split("?")[0]` is now spread 4 times along the code. Not sure but maybe we should merge the use of it

src/webapp/pages/settings/SettingsPage.tsx
```diff
@@ -232,6 +235,17 @@ export const SettingsPage: React.FC = () => {
                                     </Button>
                                 </GridForm>
                             </SubContainer>
+                            <InlineInputSave
+                                title={i18n.t("Matomo Container Tag URL")}
+                                label={i18n.t("Url")}
+                                value={analyticsConfig?.codeUrl ?? ""}
+                                onChange={value => updateAnalyticsConfig({ codeUrl: value })}
+                                onUpdate={value => {
```
p3rcypj: Both useCallbacks

src/domain/entities/AnalyticsConfig.ts
```diff
@@ -0,0 +1,3 @@
+import { Maybe } from "../../types/utils";
+
+export type AnalyticsConfig = { codeUrl: Maybe<string> };
```
p3rcypj: I assume that the creation of a separate `AnalyticsConfig` outside the `Config` entity is the effect of actually not having the representation of the "constants" we have as domain settings, by the misuse of retrieving and setting them on the `ConfigRepository` (which indeed is not the correct way of doing it). But at the same time, we're declarating Matomo Analytics and Google Analytics as separate things, something that in my opinion, I would merge. So we could delete the misuse of getting and setting `googleAnalyticsCode` on `ConfigRepository` also, and add `gaCode` here on `AnalyticsConfig`. What do you think?

src/webapp/CompositionRoot.ts
```diff
@@ -116,6 +125,9 @@ export async function getCompositionRoot(instance: Instance) {
         analytics: getExecute({
             sendPageView: new SendPageViewUseCase(analyticsRepository, configRepository),
         }),
+        matomo: getExecute({
```
p3rcypj: In my opinion, I would add `trackView` under analytics as I see both GA and Matomo as analytics.

src/webapp/CompositionRoot.ts
```diff
@@ -116,6 +125,9 @@ export async function getCompositionRoot(instance: Instance) {
         analytics: getExecute({
             sendPageView: new SendPageViewUseCase(analyticsRepository, configRepository),
```
p3rcypj: Here we could use `analyticsConfigRepository` instead of using `configRepository`.

src/webapp/CompositionRoot.ts
```diff
@@ -61,6 +66,8 @@ export async function getCompositionRoot(instance: Instance) {
     const actionRepository = new ActionDefaultRepository(config);
     const landingPageRepository = new LandingNodeDefaultRepository(config.storageClient);
     const analyticsRepository = new GoogleAnalyticsRepository();
```
p3rcypj: As we have now both GA and Matomo, we could add specificity now

src/webapp/pages/home/HomePage.tsx
```diff
@@ -263,3 +256,49 @@ function useRedirectOnSinglePrimaryNode(
...
+        title: "Homepage - Available Home Pages",
+    };
+    compositionRoot.analytics.sendPageView(viewOptions);
+    compositionRoot.matomo.trackView(viewOptions);
```
p3rcypj: Actually here, I see defined "track with Matomo" as a separate use case from "track with Google Analytics". I would say the actual use case is "track view or send page view". I think this is something we would have on the `useCase` and call each one as a part of domain logic decision. But we can let this as it is if you consider isn't the case, as I recognize this more as a personal opinion. So feel free of whatever you decide.

src/webapp/pages/home/HomePage.tsx
```diff
@@ -17,9 +17,9 @@ import { Cardboard } from "../../components/card-board/Cardboard";
 import { BigCard } from "../../components/card-board/BigCard";
 import { goTo } from "../../utils/routes";
 import { defaultIcon, defaultTitle } from "../../router/Router";
-import { useAnalytics } from "../../hooks/useAnalytics";
 import { Maybe } from "../../../types/utils";
 import i18n from "../../../utils/i18n";
+import { CompositionRoot } from "../../CompositionRoot";
 
 export const HomePage: React.FC = React.memo(() => {
```
p3rcypj, line 25: The state management is growing moment by moment. Not something I would do in this PR, but I would add a TODO comment encouraging the next developer to extract state into a `useHomePage` hook.

src/domain/usecases/TrackMatomoViewUseCase.ts
```diff
@@ -0,0 +1,23 @@
...
+import { AnalyticsEvent } from "../entities/AnalyticsEvent";
+import { AnalyticsConfigRepository } from "../repositories/AnalyticsConfigRepository";
+
+export class TrackMatomoViewUseCase {
+    constructor(
+        private configRepository: AnalyticsConfigRepository,
+        private matomoAnalytics: MatomoAnalyticsRepository
+    ) {}
+
+    public async execute(options?: { location?: string; title?: string }) {
```
p3rcypj: We have both location and title as optional, which also has it's own handling logic whenever they're not defined (on the use case). Something that on other apps I see it as a presentation thing as each component could send an event with specific (or not) location / title (for example in UHCPW), here on HomePage app we're defining location and title each time we track a page view (and events in the future) as it is a generic app. Do you think we should search for location and title optional args along the code and remove the use optional?

p3rcypj (changes requested):
Thank you, @eperedo, for your work! Analytics isn't an easy thing to deal with, and for me, it's always a big mental effort. I apologize for the delayed comments; I couldn't put words to the things I've seen over the several days I've reviewed the PR. Reading the comments will suggest you the reason why it was difficult for me to express myself, as I think most of the comments are my own personal opinion and not a clear thing to do, even if I think to myself that some of them really are.
Thank you very much for your work @eperedo! 💪🏼

p3rcypj:
I've tested the PR functionally, and I've noticed that two page view events were being sent: one that includes the app name and the page name, and another with just the page name. Additionally, after another round of functional review, I've noticed that several (seven times) short lifespan events were appearing. At the same time, I've had the strange feeling of seeing a lot of events that didn't have to occur. Almost one of the several reasons analytics tracking it's a big mental effort to manage, as useEffects send views out of nowhere.

This same issue happened to me back when I was implementing Google Analytics. I don’t recall the exact technical fix I applied, but I remember it had to do with how the useEffect dependencies were triggering — possibly due to shared state or multiple useEffects overlapping in a way that caused the event to fire more than once. It might be worth checking if there are any effects being triggered more than once due to the current dependency setup or state transitions, as the Homepage App has somewhat spaghetti state management.

I just wanted to flag it in case it helps. ✌🏼

p3rcypj:
By the time I was adding the comments (prior to it actually), I saw that you had already pushed another commit adding the page views for 'AdditionalComponents.' Now I've focused on expressing myself with tangible words about the code and some weird behaviors I've encountered with the events. However, another functional review round should be expected. @eperedo @adrianq

p3rcypj (approved):
Works as a charm! And the bug has been completely fixed. As there are a lot of "Homepage" on a lot of our apps, I would really encourage to add the prefix "HPA -". Thanks @eperedo for your hard work! Some minor comments are pending, but we can ignore/add a simple `//TODO:` comment in the code for future address

PR #45, opened 2025-04-30
Notifications management

gqcorneby:
### :pushpin: References

-   **Issue:** Closes #8698n31xh [[Homepage App] Implement notifications system](https://app.clickup.com/t/8698n31xh)

### :memo: Implementation
- updated Notification Entity
- fetch, add, delete notification access
- fetch, add, edit, delete notifications
- edit sharing settings per notification
- edit notification dialog updates
    - add preview and more information about wildcard field
    - fixed preview for edit landing page
- redirect to / on unauthorized routes 

NOTE:
- copied behavior from landing page: when saving a notification, the notification is marked as read by current user

Tested using 2 users

https://github.com/user-attachments/assets/3c11adb6-bc30-4dbe-8c2a-14c0eead3021

image

https://github.com/user-attachments/assets/f3df0b7d-dbe2-4cb2-bb5f-84dfcd1a1bd6

### :fire: Testing
https://dev.eyeseetea.com/who-dev-41/api/apps/Homepage-App---Notifs/index.html#/settings


ifoche:
Task linked: [CU-8698n31xh [Homepage App] Implement notifications system](https://app.clickup.com/t/8698n31xh)

src/data/repositories/NotificationDefaultRepository.ts
```diff
@@ -47,6 +51,18 @@ export class NotificationDefaultRepository implements NotificationRepository {
...
 
+    private _delete(notifications: Notification[]): FutureData<void> {
+        return Future.fromPromise(
+            this.storageClient.removeObjectsInCollection(
+                notificationsDataStore,
+                notifications.map(notification => notification.id)
+            )
+        ).flatMapError(error => {
+            console.error(`Notification (remove): ${error}`);
+            return Future.error(i18n.t("An error has occurred deleting notifications"));
```
tokland: The `error` object may contain useful information about what went wrong, we may include it also in the future error string.

src/domain/entities/Notification.ts
```diff
@@ -1,13 +1,37 @@
...
+                    {
+                        id: user.id,
+                        name: user.name,
+                        date: new Date(),
+                    },
+                ],
+            });
+        }
+    }
+    isRead(user: User): boolean {
```
tokland: We may add a blank line between methods. Also, very minor, maybe it's more declarative: `isReadBy(user: User)`?

src/domain/usecases/GetNotificationsUseCase.ts
```diff
@@ -0,0 +1,17 @@
...
+
+import { NotificationRepository } from "../repositories/NotificationRepository";
+import { FutureData } from "../types/Future";
+import { Notification } from "../entities/Notification";
+
+export class GetNotificationsUseCase {
+    constructor(private notificationRepository: NotificationRepository) {}
+
+    public execute(): FutureData<Notification[]> {
+        return this.notificationRepository.list(null).map(notifications =>
```
tokland: I know `notificationRepository.list` is not strictly related to this PR, but it's a recent method, let's align with our typical approach: we may use an object with optional/required props instead of Maybe and call it `list({})`. You may also allow `list()`, but that's a minor bonus. As you see fit.

src/webapp/components/user-notification/NotificationDetailsDialog.tsx
```diff
@@ -0,0 +1,85 @@
...
+    title: false,
+    dataSharing: false,
+    publicSharing: false,
+    externalSharing: false,
+    permissionPicker: false,
+};
+
+const wildCardOptions: DropdownItem[] = Object.values(NotificationWildcard).map(wildcard => ({
+    value: wildcard,
+    text: wildcard,
```
tokland: Not sure if it's a problem, but just as a note, the notification wildcards are not translatable, right?

src/webapp/components/user-notification/useNotificationDetailsDialog.ts
```diff
@@ -0,0 +1,89 @@
...
+        notification,
+        onContentChange,
+        onSharingChanged,
+        onWildcardChange,
+        searchSharing,
+        sharingMeta: generateSharingMeta(notification),
+    };
+};
+
+function newNotification(): NotificationAttrs {
```
tokland: NotificationAttrs is the object that Notification.create takes, and since that's a public method, it's ok to export it.

BUT: we should not use it from the presentation layer. 

Either we work directly with "Notification" or say, no, "here in the React component I don't really want a Notification, with its particular validations and such, I want a plain object". That's perfectly valid, and then what we do is create a `NotificationViewModel` (which can be `= NotificationProps`, if that's the case, but that's written just in once place), but we make explicit that it's a view model that may or may not mimic the entity props. (separation of concerns)

src/webapp/components/user-notification/useNotificationDetailsDialog.ts
```diff
@@ -0,0 +1,89 @@
...
+    return {
+        id: generateUid(),
+        content: "",
+        recipients: { users: [], userGroups: [], wildcard: NotificationWildcard.ALL },
+        readBy: [],
+        createdAt: new Date(),
+    };
+}
```
tokland, line 72: We can move this to the entity (`static initial(): Notification` or similar)

src/webapp/components/user-notification/useNotificationDetailsDialog.ts
```diff
@@ -0,0 +1,89 @@
...
+
+const mapRecipientToSharingRules = (recipients: NamedRef[]): SharingRule[] => {
+    return recipients.map(({ id, name }) => ({ id, displayName: name, access: "--------" }));
+};
+const mapSharingRulesToRecipient = (sharingRules: SharingRule[]): NamedRef[] => {
+    return sharingRules.map(({ id, displayName }) => ({ id, name: displayName }));
+};
+
+const generateSharingMeta = (notification: NotificationAttrs) => ({
```
tokland: same, we would use Notification (or NotificationViewModel)

src/webapp/pages/settings/CreateButton.tsx
```diff
@@ -9,10 +9,11 @@ import {
     LandingPageEditDialog,
     LandingPageEditDialogProps,
 } from "../../components/landing-page-edit-dialog/LandingPageEditDialog";
-import { Description, OpenInBrowser } from "@material-ui/icons";
+import { Description, Message, OpenInBrowser } from "@material-ui/icons";
 
 export interface CreateButtonProps {
     landings: Maybe<LandingNode[]>;
+    newNotification: () => void;
```
tokland: "newNotification" conveys a substantive, an object. But this is a callback that we will call when we have new notifications, right? maybe `onNewNotification`?

src/webapp/pages/settings/useNotifications.ts
```diff
@@ -0,0 +1,104 @@
...
+    useEffect(() => {
+        setNotifDetailsDialog(props =>
+            props
+                ? {
+                      ...props,
+                      isLoading: isLoading,
+                  }
+                : undefined
+        );
+    }, [isLoading]);
```
tokland, line 127: useEffect blocks are anonymous. We have two approaches, either move it to a function (and then it has a name) or add a comment explaining what it does.

tokland (commented):
@gqcorneby I know the PR is still not finished, but I think a first review won't hurt, this are mostly orthogonal notes that apply in any case.

src/webapp/components/user-notification/NotificationDetailsDialog.tsx
gqcorneby: Yes, the wildcard is an item from a fixed selection.

i18n/en.pot
```diff
@@ -47,13 +47,29 @@ msgid ""
 "interrupted."
 msgstr ""
 
-msgid "An error has occurred fetching notifications"
+msgid ""
+"An error has occurred fetching notifications\n"
+"{{error}}"
```
tokland: [no need to change] when the interpolated string is so clearly outside the translation literal, we typically keep only the text ("An error has occurred fetching notifications") and do the interpolation in the code. This way we keep clean literals to be translated.

src/data/repositories/NotificationDefaultRepository.ts
```diff
@@ -33,37 +38,58 @@ export class NotificationDefaultRepository implements NotificationRepository {
         return this._delete(notifications);
     }
 
+    public getConfig(): FutureData<NotificationConfig> {
+        return this._getConfig();
+    }
+
+    public saveConfig(config: NotificationConfig): FutureData<void> {
```
tokland: Not sure Jorge would approve this getConfig and saveConfig. This is a NotificationRepository, not a NotificationConfigRepository. As we already have a ConfigRepository, I guess it makes sense to add `config: NotificationConfig` there and `ConfigRepository.get` would be the responsible for setting the value. Confirm with @xurxodev 

[edit] I see that we have also save of that config. So in that case, NotificationConfigRepository may make more sense. Anyway, check with Jorge.

src/domain/entities/generic/Struct.ts
```diff
@@ -32,11 +34,26 @@ export function Struct<Attrs>() {
...
+
+        static tryCreate<U extends Base>(this: new (attrs: Attrs) => U, attrs: Attrs): Future<Error, U> {
+            const instance = new this(attrs);
+            const validationResult = instance.validateOrTransform();
+
+            return validationResult.match({
+                success: value => Future.success(value),
+                error: message => Future.error(message),
+            });
+        }
```
tokland, line 50: Cool. 

[nothing do do, just a note] It's cheating a bit to return a Future instead of an Either (as we are really doing anything asynchronous here), but it makes sense to simplify its usage in a Future chain).

src/domain/usecases/SaveNotificationConfigUseCase.ts
```diff
@@ -0,0 +1,23 @@
...
+                    ...config.permissions,
+                },
+            }))
+            .flatMap(updatedConfig => this.notificationRepository.saveConfig(updatedConfig));
+    }
+}
+
+type ShallowPartial<T> = {
+    [K in keyof T]?: T[K] extends object ? { [P in keyof T[K]]?: T[K][P] } : T[K];
+};
```
tokland: It's a generic abstraction, it can be moved to src/types/utils.ts

src/domain/usecases/SaveNotificationConfigUseCase.ts
```diff
@@ -0,0 +1,23 @@
+import { NotificationRepository } from "../repositories/NotificationRepository";
+import { FutureData } from "../types/Future";
+import { NotificationConfig } from "../entities/Notification";
+
+export class SaveNotificationConfigUseCase {
+    constructor(private notificationRepository: NotificationRepository) {}
+
+    public execute(config: ShallowPartial<NotificationConfig>): FutureData<void> {
```
tokland: We can create a type alias for "ShallowPartial<NotificationConfig>" so the signature is cleaner.

src/webapp/components/dropdown-with-desc/DropdownDesc.tsx
```diff
@@ -0,0 +1,129 @@
...
+                renderValue={value => {
+                    const option = options.find(opt => opt.value === value);
+                    return option ? (
+                        <>
+                            {option.text} - <ItemDescription>{option.desc}</ItemDescription>
+                        </>
+                    ) : (
+                        ""
+                    );
+                }}
```
tokland: I think that too much code to inline in a prop value, can we move it to a component?

src/webapp/components/user-notification/NotificationListTable.tsx
```diff
@@ -6,17 +6,54 @@ import {
...
+                        userAccesses: userAccesses || notification.permissions.userAccesses,
+                        userGroupAccesses: userGroupAccesses || notification.permissions.userGroupAccesses,
+                        publicAccess: publicAccess || notification.permissions.publicAccess,
+                    },
+                };
+                await saveNotifications([updatedNotification]);
+            },
+            onClose: () => setPermissionNotificationId(undefined),
+        };
+    }, [notifications, permissionNotificationId, saveNotifications]);
```
tokland: Quite long, can we move to a function?

src/webapp/hooks/useUserNotifications.ts
```diff
@@ -26,21 +26,31 @@ export function useUserNotifications(props: useUserNotificationProps) {
...
+                                    .save([notification], appContextProps.currentUser)
+                                    .toPromise();
+                                closeNotificationDialog(notification);
+                            },
+                        }))
+                    );
+                } else {
+                    continueLoading();
+                }
+            } catch {
```
tokland: that useEffect body is quite large, can we create a function and call it?

src/webapp/pages/settings/useConfig.ts
```diff
@@ -83,6 +84,27 @@ export function useConfig(): useConfigPloc {
...
+                userGroupAccesses:
+                    settingsPermissions?.userGroups?.map(ref => ({
+                        ...ref,
+                        access: "rw----",
+                    })) ?? [],
+            },
+            onChange: updateSettingsPermissions,
+        }),
+        [settingsPermissions, updateSettingsPermissions]
+    );
```
tokland, line 122: Similar to before, when we have a block with relevant logic, it's nice to abstract to function. The idea is to reduce the body of the hook, so we can see in a glance what it's doing.

src/domain/entities/Notification.ts
```diff
@@ -27,11 +30,51 @@ export class Notification extends Struct<NotificationAttrs>() {
...
 
+    canView(user: User) {
+        return this.checkPermissionAccess(user, "r");
+    }
+
+    canEdit(user: User) {
+        return this.checkPermissionAccess(user, "rw");
+    }
+
+    private checkPermissionAccess(user: User, accessType: "r" | "rw"): boolean {
```
tokland, line 46: [nothing to do] I am not a big fan of using the dhis2 access strings in the domain. It's ok to use D2 specifics when they look good, this string is not nice to use :) But I see we already have in the domain publicAccess and access props as strings, so we will use it. But for new projects, i think it's nicer to have more declarative object types for accesses types.

tokland (commented):
In-line:

src/data/repositories/NotificationDefaultRepository.ts
gqcorneby: Got it! Initial plan was just permissions as discussed last week but I decided to save it as a config instead so it's expandable just in case. I'll create a NotificationConfigRepository.

src/webapp/pages/settings/useConfig.ts
```diff
@@ -6,6 +6,22 @@ import { User } from "../../../domain/entities/User";
 import { Maybe } from "../../../types/utils";
 import { LandingNode, updateLandings } from "../../../domain/entities/LandingNode";
 import i18n from "../../../utils/i18n";
+import { useAccessPermissionsDialog } from "./useAccessPermissionsDialog";
+
+type useConfigPloc = {
```
tokland: type, UpperCase

src/types/utils.ts
```diff
@@ -20,6 +20,10 @@ export type RecursivePartial<T> = {
         : T[P];
 };
 
+export type ShallowPartial<T> = {
```
tokland, line 28: Type helpers are typically hard to grasp by name or implementation, can we add a doc (simple) with explanation and some use case?

tokland (approved):
I only spotted some very minor, non-critical issues, so approving.

ifoche (approved):
thanks so much @gqcorneby . Working like a charm

PR #47, opened 2025-05-29
Feature: translatable notifications

gqcorneby:
### :pushpin: References

-   **Issue:** Closes #86996ww4z [[Homepage App] Notification translations implementation](https://app.clickup.com/t/86996ww4z)

### :memo: Implementation
- to handle backwards compatibility, we'll be adding a new prop to the notification datastore model `translations: { es: ".."}`
- add data store notification model
- update notification entity content from string to TranslatableText
- import/export translations handling

https://github.com/user-attachments/assets/7ee5677c-13f3-48e8-963b-20e5f2d4fe48


ifoche:
Task linked: [CU-86996ww4z [Homepage App] Notification translations implementation](https://app.clickup.com/t/86996ww4z)

ifoche (changes requested):
@gqcorneby it seems like we don't have translations for Frech or Spanish in the notifications interface. The "Close" and "Okay" buttons show in English. Could we add them?
2025-06-06-232436_hyprshot

I would use "Cerrar" and "Aceptar" in Spanish
I would use "Fermer" and "OK" in French

Thank you!

tokland:
@gqcorneby The code in this PR is related to the coding dojo task on export/import and Clean Architecture, right? I'm removing myself as a reviewer for now — please re-add me when it's ready for review.

(This assumes it's okay to delay approval of PRs related to the Clean Architecture refactor. Please confirm with your PM)

src/data/repositories/NotificationDefaultRepository.ts
```diff
@@ -2,14 +2,24 @@ import _ from "lodash";
...
 import { Instance } from "../entities/Instance";
 import { DataStoreStorageClient } from "../clients/storage/DataStoreStorageClient";
 import { notificationNamespaceKeys, notificationsNamespace } from "../clients/storage/Namespaces";
 import { StorageClient } from "../clients/storage/StorageClient";
 import { Maybe } from "../../types/utils";
 import i18n from "../../utils/i18n";
 
+type DataStoreNotification = Omit<NotificationAttrs, "content"> & {
+    content: string;
+    translations?: Record<string, string>;
```
tokland: There are many references to Record as translations or terms. It makes sense to create some aliases (not necessarily just one, but group them when they refer to the same structure).

src/domain/usecases/ExportNotificationsTranslationsUseCase.ts
```diff
@@ -0,0 +1,42 @@
...
+import { NotificationRepository } from "../repositories/NotificationRepository";
+import { UseCase } from "./UseCase";
+import JSZip from "jszip";
+import FileSaver from "file-saver";
+import { TranslationService } from "./helpers/TranslationService";
+import { Future, FutureData } from "../types/Future";
+import { fromPromise } from "../../data/api-futures";
+import { Notification } from "../entities/Notification";
+
+export class ExportNotificationsTranslationsUseCase implements UseCase {
```
tokland, line 11: As we discussed in the Coding Dojo (this PR was written beforehand), we'll follow Jorge's advice and not create repositories or use cases for export/import. Instead, we'll use the standard Get and Set use cases, build the file, and handle the download in the presentation layer.

src/domain/usecases/ExportNotificationsTranslationsUseCase.ts
```diff
@@ -0,0 +1,42 @@
...
+    constructor(private notificationRepository: NotificationRepository) {}
+
+    public execute(ids: string[]): FutureData<void> {
+        return this.fetchAndValidateNotifications(ids).flatMap(this.exportTranslations);
+    }
+
+    private fetchAndValidateNotifications(ids: string[]): FutureData<Notification[]> {
+        return this.notificationRepository.list({ ids }).flatMap(notifications => {
+            if (!notifications || notifications.length === 0) {
```
tokland: Is that the correct logic? What if ids was empty? Maybe we want to check that we have as many notifications as ids.length ? Should this "error" check being done in the repository?

src/domain/usecases/ImportNotificationsTranslationsUseCase.ts
```diff
@@ -0,0 +1,43 @@
...
+        return this.notificationRepository.list({ ids }).flatMap(notifications => {
+            if (!notifications || notifications.length === 0) {
+                return Future.error("Unable to load notification.");
+            }
+            return Future.success(notifications);
+        });
+    }
+
+    private setTranslations =
+        (language: string, terms: Record<string, string>) =>
```
tokland: We generally prefer not to mix regular class methods with arrow function instance properties. I assume you're using an arrow function here to allow point-free style references, which makes sense from a functional programming perspective. I remember you asking me something about it, but in this particular scenario, for readability — especially for those less familiar with FP patterns— I’d suggest rewriting this as a standard method. The gain of using a curry function here is very small.

 ```
private setTranslations(language: string, terms: Record<string, string>, notifications: Notification[]): Notification[] { ... }
```

File: src/domain/usecases/ImportNotificationsTranslationsUseCase.ts [hunk 5]
```diff
@@ -0,0 +1,43 @@
...
+                const content = setTranslationValue({
+                    item: notification.content,
+                    language,
+                    term: terms[notification.content.key],
+                });
+                return Notification.create({ ...notification, content });
+            });
+        };
+
+    private saveTranslatedNotifications = (notifications: Notification[]): FutureData<number> => {
```
tokland: same, I'd put a method + bind in the caller.

src/webapp/components/notifications/NotificationListTable.tsx
```diff
@@ -15,17 +17,23 @@ import { PermissionsDialog, PermissionsDialogProps } from "../permissions-dialog
...
     saveNotifications: (notifications: NotificationViewModel[]) => Promise<void>;
     isLoading: boolean;
+    fetchNotifications: () => Promise<void>;
 };
 
 export const NotificationListTable: React.FC<NotificationListTableProps> = props => {
-    const { notifications, onEditNotification, deleteNotifications, saveNotifications, isLoading } = props;
+    const { notifications, onEditNotification, deleteNotifications, saveNotifications, isLoading, fetchNotifications } =
+        props;
+    const translationImportRef = useRef<ImportTranslationRef>(null);
```
tokland, line 34: I don't have time to delve into it, but usually, using a ref is considered a code smell. Granted, sometimes it's the only way to make everything work. So, just as a note: if we can do it with the typical useCallback, great. If that's not possible, then it's fine to keep the useRef.

src/webapp/components/notifications/NotificationListTable.tsx
```diff
@@ -151,11 +202,24 @@ function buildTableActions(props: BuildTableActionProps): TableAction<Notificati
...
             },
         },
+        {
+            name: "export-translations",
+            text: i18n.t("Export JSON translation"),
+            icon: <Icon>translate</Icon>,
+            multiple: true,
+            onClick: async (ids: string[]) => {
+                await exportTranslations(ids);
+            },
```
tokland, line 212: `onClick: exportTranslations` works or we really need to do the awaiting within?

src/webapp/components/notifications/UserNotificationDialog.tsx
```diff
@@ -17,7 +19,7 @@ export const UserNotificationDialog: React.FC<UserNotificationDialogProps> = pro
             fullWidth={true}
             onSave={onConfirm}
         >
-            <NotificationContent content={notification.content} />
+            <NotificationContent content={translate(notification.content)} />
```
tokland, line 22: cool

src/domain/usecases/helpers/TranslationService.ts
```diff
@@ -0,0 +1,38 @@
+import _ from "lodash";
+import { TranslatableText } from "../../entities/TranslatableText";
+
+//{lang: {key: translatedText}}
+type Language = string;
+type Translations = Record<Language, Record<string, string>>;
+
+export class TranslationService {
+    static extractTranslations<T>(entities: T[]): Translations {
```
tokland: We should be more specific about the types of entities we're dealing with to improve readability and predictability. One idea is to accept entities that implement a particular interface—for example, one with a method like getTranslationKeys, depending on the context. The JavaScript get-all approach, working on objects of unknown shape, can quickly become unmanageable.

src/domain/usecases/helpers/TranslationService.ts
```diff
@@ -0,0 +1,38 @@
...
+            .groupBy("lang")
+            .mapValues(array => _.fromPairs(array.map(({ key, value }) => [key, value])))
+            .value();
+
+        return { ...translatedStrings, en: referenceStrings };
+    }
+
+    private static isTranslatableText(obj: unknown): obj is TranslatableText {
+        return (
+            typeof obj === "object" && obj !== null && "key" in obj && "referenceValue" in obj && "translations" in obj
```
tokland: Needing this is also a code smell (related to my previous comment).

tokland (commented):
In-line comments:

(I may have not the full picture for some of comments, apply them as you see fit for the current state of the project)

src/domain/usecases/helpers/TranslationService.ts
gqcorneby: will revert it to entity specific functions for now "getLandingNodeTranslations" etc. Will implement entities implementing an interface in the coding dojo refactor

src/domain/usecases/ExportNotificationsTranslationsUseCase.ts
gqcorneby: In the repository, there's a check that should return all notifications if ids is empty. I guess I can change the error message to "No notifications to export"?

```
private isInIdList(notification: Notification, ids?: string[]): boolean {
        return !ids || ids.length === 0 || ids.includes(notification.id);
    }
```

src/webapp/components/notifications/NotificationListTable.tsx
gqcorneby: attempted to refactor

src/domain/usecases/ExportNotificationsTranslationsUseCase.ts
tokland: > error message to "No notifications to export"

Yes, that's ok. About:

`if (!notifications || notifications.length === 0) {`

If we trust the type `(parameter) notifications: Notification[]`, there is no need to check for `!notifications`

src/domain/usecases/helpers/TranslationService.ts
```diff
@@ -1,38 +1,31 @@
...
+    return buildTranslationMap(texts);
+}
 
-        return { ...translatedStrings, en: referenceStrings };
-    }
+function buildTranslationMap(texts: TranslatableText[]): TranslationMap {
+    const referenceStrings = _.fromPairs(texts.map(({ key, referenceValue }) => [key, referenceValue]));
+    const translatedStrings = _(texts)
+        .flatMap(({ key, translations }) => _.toPairs(translations).map(([lang, value]) => ({ lang, key, value })))
+        .groupBy("lang")
```
tokland: not type-safe, let's use an arrow function instead.

src/webapp/components/action-list-table/ActionListTable.tsx
```diff
@@ -366,18 +369,18 @@ export const ActionListTable: React.FC<ActionListTableProps> = props => {
...
-        [openImportDialog, translationImportRef]
+        [openImportDialog, startImport]
     );
 
     return (
         <PageWrapper>
             {dialogProps && <ConfirmationDialog isOpen={true} maxWidth={"xl"} {...dialogProps} />}
 
-            <ImportTranslationDialog type="action" ref={translationImportRef} onSave={handleTranslationUpload} />
+            {ImportTranslationDialog}
```
tokland: The consensus in the community seems to be that hooks should not return components or views; they should concern themselves only with logic. 

I see that this pattern is used at least 3 times in the app. Approaches:

- Let's go with it, with just a note in the hook.
- Revert to the old code: `<ImportTranslationDialog type="action" ref={translationImportRef} onSave={handleTranslationUpload} />`
- A typical pattern: make the hook return an object that can be directly used as props of the view component: 
```
const { startImport, importTranslationDialogOptions } = useImportTranslationDialog({
...
<ImportTranslationDialog {...importTranslationDialogOptions} />
```

src/webapp/components/import-translation-dialog/ImportTranslationDialog.tsx
```diff
@@ -1,109 +1,73 @@
...
-                try {
-                    const file = event.target.files[0];
-                    if (!file) throw new Error("No file received on upload");
-
-                    const text = await file.text();
-                    const json = JSON.parse(text);
-                    const terms = _.pickBy(json, value => _.isString(value));
+export type ImportTranslationDialogProps = {
+    type: "action" | "landing-page" | "notification";
+    onSave: (key: string | undefined, lang: string, terms: Record<string, string>) => void | Promise<void>;
```
tokland: [No need to change] For the record, `key: string | undefined, lang: string, terms: Record<string, string>` is a group of data that's used more than once. That typically means that there is an object type to be created that "unifies" them as a single concept.

src/webapp/components/import-translation-dialog/ImportTranslationDialog.tsx
```diff
@@ -1,109 +1,73 @@
...
-            <React.Fragment>
-                <input
-                    type="file"
-                    name="file"
-                    accept={"application/json"}
-                    ref={inputRef}
-                    onChange={onFileUpload}
-                    style={{ display: "none" }}
-                />
+        if (!selectedLang || !terms) {
```
tokland: `!terms` Is that check necessary? it's an object, it will be always false.

src/webapp/components/import-translation-dialog/useImportTranslationDialog.tsx
```diff
@@ -0,0 +1,68 @@
...
+    );
+
+    const importDialog = (
+        <>
+            <input
+                type="file"
+                accept="application/json"
+                onChange={onFileUpload}
+                ref={fileInputRef}
+                style={{ display: "none" }}
```
tokland: not important, but as a general rule, and since we have all the other props immutable, we can make style also immutable. We can move it as an object in the root level.

tokland (commented):
In-line:

src/webapp/components/action-list-table/ActionListTable.tsx
gqcorneby: reverted

src/webapp/pages/settings/SettingsPage.tsx
```diff
@@ -1,359 +1,328 @@
...
+function buildActionList(actions: Action[], translate: TranslateMethod): DropdownItem[] {
+    return actions.map(({ id, name }) => ({ value: id, text: translate(name) }));
+}
 
-export const GridForm = styled.div`
-    display: grid;
-    grid-template-columns: 2fr 1fr;
-    gap: 20px;
-`;
+const languages = [
```
tokland: How come we need this list of languages now? In any case, it probably should not be hardcoded. Do we want to support any language? If so, there are npm packages with the ISO codes. Or do we want to support the DHIS2 locales? In that case, we'd get the list from the instance in some repo.

gqcorneby: Sorry about that! I'm not sure how this was running for me locally yesterday. This was a mistake while reverting the import dialog.

tokland (approved):
Approved!

(As we discussed, the option of removing useRef was optimal, but the gain is not significant enough to keep the PR stalled; it's good enough)

ifoche (approved):
thanks @gqcorneby

PR #51, opened 2025-08-08
Fix/limit all wildcard

gqcorneby:
### :pushpin: References

-   **Issue:** Closes #869a2gztq [Limit "All" wildcard to platform admins](https://app.clickup.com/t/869a2gztq)

### :memo: Implementation
- limiit `ALL` wildcard to superAdmin
- show create Notif only if has Access

**Left (superAdmin)** - **Right (user)**

https://github.com/user-attachments/assets/20579586-e325-49e5-9854-e3c23bce203e

### DIFF
--- a/i18n/ar.po
+++ b/i18n/ar.po
@@ -1,12 +1,13 @@
 msgid ""
 msgstr ""
 "Project-Id-Version: i18next-conv\n"
-"POT-Creation-Date: 2025-06-06T19:59:41.289Z\n"
+"POT-Creation-Date: 2025-08-14T11:48:09.626Z\n"
 "PO-Revision-Date: 2025-06-06T19:59:41.289Z\n"
 "MIME-Version: 1.0\n"
 "Content-Type: text/plain; charset=UTF-8\n"
 "Content-Transfer-Encoding: 8bit\n"
-"Plural-Forms: nplurals=6; plural=(n==0 ? 0 : n==1 ? 1 : n==2 ? 2 : n%100>=3 && n%100<=10 ? 3 : n%100>=11 && n%100<=99 ? 4 : 5)\n"
+"Plural-Forms: nplurals=6; plural=(n==0 ? 0 : n==1 ? 1 : n==2 ? 2 : n%100>=3 "
+"&& n%100<=10 ? 3 : n%100>=11 && n%100<=99 ? 4 : 5)\n"
 
 msgid "There are pending migrations"
 msgstr ""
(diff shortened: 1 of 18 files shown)


ifoche:
Task linked: [CU-869a2gztq Limit "All" wildcard to platform admins](https://app.clickup.com/t/869a2gztq)

ifoche (changes requested):
Hi @gqcorneby . Please find here some review from Codex I find very interesting:

Review
The branch hides the “All” wildcard option for non‑super‑admin users.
This is implemented through wildCardOptions, which filters allWildCardOptionInfo by the hide function and only exposes the “All” option when isSuperAdmin(user) is true. UI components (NotificationDetailsDialog and NotificationListTable) rely on this filtered list, so non‑admins no longer see “All”.

Potential issues

- The restriction is enforced only in the UI. A user with notification‑write access could still craft a request directly to the datastore and set wildcard: "ALL" because neither SaveNotificationsUseCase nor the repository validates the value. A server‑side or domain‑level guard would make the restriction harder to bypass.

- Editing an existing “ALL” notification as a non‑admin results in the dropdown rendering empty (no matching option). If the user saves without changing the value, the notification remains “ALL”, so the editor might not realize the wildcard hasn’t changed.

Overall, the patch fulfills the requirement for the UI, but further validation may be needed to fully enforce the rule and to handle editing of existing “ALL” notifications gracefully.

Could you please address these potential issues? 

Thank you!

gqcorneby:
Hi @ifoche! Yes this is a good note. I will add restriction in the data/domain layer as well. Thank you!

ifoche (changes requested):
Hi @gqcorneby . Please find some additional comments still from codex, that seem to be valid:
Issues & Suggestions
- **Non‑super‑admin users cannot mark global (“ALL”) notifications as read** because validation blocks all saves, even when the user is only confirming the notification. This leaves the notification unread and it will reappear on subsequent logins

Suggested task
Allow non-admins to mark ALL notifications as read

- **SaveNotificationsUseCase does not verify edit permissions before saving**. A user with read access could craft a manual call that modifies notifications they are not allowed to edit, because only wildcard validation is performed

Suggested task
Enforce edit permissions when saving notifications


Could you please fix those?

Thank you!

gqcorneby:
Hi @ifoche! Updated. Sorry about that, I kept testing by closing notifs so I could keep testing with the same notif and did not notice the 1st issue right away.

ifoche (approved):
thanks @gqcorneby

PR #57, opened 2025-10-22
Fix: Export pages sharing settings

nshandra:
### :pushpin: References

-   **Issue:** Closes [https://app.clickup.com/t/8699phnqu](https://app.clickup.com/t/8699phnqu)

### :memo: Implementation

Updated the import and export landing page useCases to include the sharing settings data from `Namespaces.CONFIG`.
Also, the delete landing page useCase now also deletes the sharing settings instead of leaving the data in the config datastore.


MiquelAdell (approved):
tested in play and it exports and reimports sharing settings correctly

xurxodev (approved):
thanks @nshandra

This PR fixes an important functional gap by making landing page export/import/delete aware of sharing settings, ensuring that page permissions are exported, re-imported and cleaned up consistently.

I approve the PR.

For me, the implementation is correct for the current codebase constraints, but it also highlights several architectural problems that would be good to address in a dedicated refactor.

This are my thoughts about this code for the project managers like @adrianq or @MiquelAdell

## Recommendations for a refactor and not related to this PR

- **LandingNode without permissions (domain model)**  
  Today a `LandingNode` can exist without any notion of permissions; sharing is modeled in `Settings.landingPagePermissions`. That weakens the domain (a landing page without permissions is not a valid business state) and forces import/export/delete to sync two places.  This can provokes more errors in other places
  **Suggestion:** Make permissions part of the landing page aggregate: a value object owned by `LandingNode` or a closely related aggregate. Enforce the invariant "a landing page always has permissions" in the entity and its factory; keep `Settings` for cross-cutting config only.

- **Use cases depending on data-layer artifacts**  
  `ExportLandingNodesUseCase` and `ImportLandingNodesUseCase` depend on `ImportExportClient` and `PersistedLandingNode*` from the data layer, breaking the dependency rule (use cases should depend on domain contracts).  
  **Suggestion:** Introduce domain-level ports for export/import; implement them with `ImportExportClient` and `PersistedLandingNode*` in infrastructure. Use cases should work only with domain types (`LandingNode`, `LandingPagePermission`) or application DTOs.

- **Use cases coupled to UI/infrastructure (`File`)**  
  `ImportLandingNodesUseCase.execute(files: File[])` depends on the browser `File` type.  
  **Suggestion:** The use case should receive a domain-level input (DTO or parsed entity). The web layer should convert `File[]` to that input before calling the use case.

- **Use cases doing persistence shaping instead of domain logic**  
  Import and export work in terms of `PersistedLandingNode*` and manual array merging, which is persistence/data shaping.  
  **Suggestion:** Move persisted-shape responsibilities to repositories/adapters; keep use cases orchestrating domain entities and value objects, with conversion to/from storage at the edges.

- **Richer LandingNode aggregate (follow-up)**  
  Many of the above issues are addressed if `LandingNode` is the aggregate root: owns permissions, encapsulates invariants and children, and defines its export/import representation.  
  **Suggestion:** As a follow-up task, define the "complete" landin aggregate (structure + permissions + invariants) and refactor these use cases and repositories around it, with persistence and file handling in the outer layers.

- **Tests after refactor**  
  Once permissions live inside the landing page aggregate and creation is forced through factory/constructors that require permissions, we can add tests that validate the invariant "a landing node always has permissions" (e.g. unit tests on the entity/factory and integration tests on create/import flows).

adrianq:
Thanks @xurxodev and @nshandra I have created a ticket to capture your thoughts on the refactor https://app.clickup.com/t/869cc2pxj Merging this...

PR #58, opened 2025-11-04
Fix/row size

gqcorneby:
### :pushpin: References

-   **Issue:** Closes #869b1czj5 [Homepage App: Review logic for number of actions per row](https://app.clickup.com/t/869b1czj5)

### :memo: Implementation
- set default row size to be calculated based on visible actions
- add configurable landing row size

https://github.com/user-attachments/assets/51d26d1e-6264-4f17-b850-b94d01c1e750


xurxodev (changes requested):
Thanks @gqcorneby.

The funcionality works but I have a concerns about the architecture.

There is an **architectural concept error** in this app: **schemas (Codec or Zod) are used in the domain layer** to define entities. Schemas for validation/serialisation should live in the **data layer** or the **presentation layer**, not in domain. This is not introduced by this pull request.

This mistake goes back to **training app**, which this app was derived from. Defining domain entities only as types derived from schemas leads to **data and behaviour being separated and not cohesive**.

How this PR doesn't introduce this architecture problem, I only have 3 changes to request

1. Move new getUserActions to own file because mix logic using two entities (action, user).
   It's like a domain service and it's more clear if it's in a separate file.

2. Can you create unit tests for this new separated function to avoid regression bugs in the future?

3.  In **`ItemRoot`** component, the presentation logic is mixed with render logic. Please create a **custom hook** and move that logic there, returning:  `isSinglePage`, `landingRowSize`, `childrenRowSize` from the hook

Ramon-Jimenez (approved):
All good

xurxodev (approved):
Thanks @gqcorneby

PR #49, opened 2025-07-03
[coding dojo] refactor settings module

eperedo:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/86994rent

### :memo: Implementation

Check `coding-dojo.md` for implementation details

### :fire: Testing

#86994rent

PR #60, opened 2025-12-15
[refactor]: Improve homepage performance

deeonwuli:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/869b9z4qu

### :memo: Implementation
- [x] Reduce repeated API calls on homepage load
- [x] Create a repository for `Settings` to manage homepage settings 
- [x] Clean up/improve ConfigRepository and related files
- [x] Refactor code to better align with clean architecture principles

Number of requests after refactor: 66
Screenshot 2025-12-15 at 14 15 12

Number of requests before refactor: 418
Screenshot 2025-12-15 at 14 25 37

### :fire: Testing

#869b9z4qu


src/data/repositories/Dhis2ConfigRepository.ts
```diff
@@ -35,7 +29,7 @@ export class Dhis2ConfigRepository implements ConfigRepository {
     }
 
     @cache()
-    public async getUser(): Promise<User> {
+    private async getUser(): Promise<User> {
```
eperedo: Since you're doing a lot of refactor in terms of clean arch. I think we should remove this Repository and the `Config.ts` file because this entity is exposing [implementation details on the domain model](https://github.com/EyeSeeTea/home-page-app-dev/pull/49/commits/43bb66c1aebeca3af186e43656d41853bf7d4b59).

You can see more details on [this PR](https://github.com/EyeSeeTea/home-page-app-dev/pull/49). Check also the `UserRepository` implementation.

eperedo (changes requested):
Looks great. Just a note, we have this [PR](https://github.com/EyeSeeTea/home-page-app-dev/pull/49) where we did some performance and clean architecture improvements. 

I think merging both branches could be a pain because we basically did the same thing, but with different names for files, variables, etc.

So I think we should close my branch and use this one because you also fixed the problem with the app verification process.

The only thing missing in this branch is removing the `Config.ts` entity and its repository. Check my comment for further details. Let me know if you have any questions.

eperedo (changes requested):
Code looks good, but some landing pages are not showing the actions. Can you check this please?

**Version installed on dev.eyeseetea**

https://github.com/user-attachments/assets/6ec976d6-db5d-4955-94c2-a2ccce4c4a5b

**Current branch**

https://github.com/user-attachments/assets/ff3eb44f-efc3-47d2-aac2-8b810e756f77

Also if you go to the Edit modal on the settings page you can see the current actions are not there:

image

but we have actions for this landing page, this is the same landing page edit modal, but on dev:

image

deeonwuli:
> Code looks good, but some landing pages are not showing the actions. Can you check this please?
> 
> **Version installed on dev.eyeseetea**
> 
>  Homepage-App_dev_actions.mp4 
> **Current branch**
> 
>  Homepage-App_no_actions.mp4 
> Also if you go to the Edit modal on the settings page you can see the current actions are not there:
> 
> image
> but we have actions for this landing page, this is the same landing page edit modal, but on dev:
> 
> image

Fixed this by removing trailing "#/" in the launch paths of the actions and installed apps so they are better matched and immediately returning false if the installed app has no launch path. This is the current branch now:

https://github.com/user-attachments/assets/ba0c1bf1-c4f3-4581-a74e-ac2c0a8d4fe2

eperedo (changes requested):
I think we're in the right direction, but now we have a problem with some actions:

Reports app is marked as "Not installed"

image

but we have the app in the server, but in the action you can see the url is pointing to an internal url:

image

Eg:

DHIS2 url in report action: "/dhis-web-reports/index.html#/standard-report"
DHIS2 report launch path from "/api/apps" endpoint: "/dhis-web-reports/index.html"

They are different so the condition is not met. 

So I think we need to include an additional logic to ignore internal urls, if we find a hash "#" we can assume the first part of the url is the app launch path then ignore the internal url. What do you think? @deeonwuli 

Also I think we should introduce a new dropdown in the action form to select the APP (this dropdown would show all the installed apps) This would allow us to easily detect if an app is installed, and we could use the DHIS2 application text input only for the specific URL we want to redirect the user to when the action is clicked. I think we need the input from @MiquelAdell or @ifoche here?

src/data/utils/d2-api.ts
```diff
@@ -15,11 +15,12 @@ export function getD2APiFromInstance(instance: Instance): D2Api {
 
 export function isAppInstalledByUrl(launchPath: string, installedApps: InstalledApp[]): boolean {
     const isPathRelative = launchPath.startsWith("/");
-    if (!isPathRelative) return false;
+    if (isPathRelative) return true;
```
eperedo: With this change any path that starts with "/" will be considered "installed", right? So the validation if the url is in the installed apps is never executed. I think we don't need this early return, right?

```shell
/dhis-installed-app/index.html -> startsWith "/" -> mark as installed
/dhis-NOT-installed-app/index.html -> startsWith "/" -> mark as installed
```

eperedo (changes requested):

MiquelAdell:
@eperedo  I've have used this option to link to an installed app with parameters something like:
`/capture-app?orgUnit=AAABBBCC` and that would not be possible just with a drop down. To not extend this I would leave this  with a text entry and we will only need @deeonwuli to address the remaining issue.

Maybe a drop down is a good solution we can try to code for in the future but having these advanced uses in mind

Thank you both!

src/data/utils/d2-api.ts
deeonwuli: Right Eduardo. The early return has been removed. Now `isAppInstalledByUrl` will properly validate all paths (including relative ones starting with "/") against the installed apps list, instead of incorrectly marking all relative paths as installed. Thank you.

deeonwuli: Screenshot 2026-01-15 at 12 04 28

Actually @eperedo, we face another problem here. The installed apps list does not contain apps such as Messaging and User settings even though they are present in the instance. So it returns false when the `isAppInstalledByUrl` should return `true`. We could get the installed apps using this endpoint: `/dhis-web-commons/menu/getModules.action`. What do you think?

eperedo: Yes, that endpoint is including all the applications and it works the same in v42 so I think is safe to use instead the `/api/apps` endpoint

deeonwuli: I tried using that endpoint @eperedo but I ran into an issue: `dhis-web-commons/menu/getModules.action` returns 404 not found when used with d2-api `get` method because we make a call to `/api/dhis-web-commons/menu/getModules.action`. I'm proposing that we make use of `api.baseConnection.request` as we previously did but only for the apps not included in the installed apps list. That way we are not making as many calls as we did before:

```
 if (isPathRelative && !isAppInstalled) {
        try {
            const response = await api.baseConnection.request({ method: "get", url: normalizedLaunchPath }).getData();
            return response !== undefined && response !== null && response !== "";
        } catch (error: any) {
            return false;
        }
    }
    return isAppInstalled;
```

eperedo: Yes. I think a mix could be a good solution. It should not be a big impact on performance because there's only 4 apps missing in the `/api/apps` endpoint

deeonwuli: All done now @eperedo

eperedo (approved):
Looks good, thanks @deeonwuli!

Just as a note: Testing from localhost against the dev API may trigger CORS errors, potentially causing apps to incorrectly report a 'not installed' status.

PR #61, opened 2026-02-05
[fix]: Homepage refresh

deeonwuli:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/869azwn9r

### :memo: Implementation
- [x] Removed the `isLoadingLong` state variable and its associated 8-second timeout that caused the homepage to refresh 

https://github.com/user-attachments/assets/f09200a1-c5e6-40a2-9e77-f1d604046fbb

### :fire: Testing

#869azwn9r

### DIFF
--- a/i18n/ar.po
+++ b/i18n/ar.po
@@ -1,7 +1,7 @@
 msgid ""
 msgstr ""
 "Project-Id-Version: i18next-conv\n"
-"POT-Creation-Date: 2025-12-15T12:49:42.172Z\n"
+"POT-Creation-Date: 2026-02-05T16:47:42.188Z\n"
 "PO-Revision-Date: 2025-06-06T19:59:41.289Z\n"
 "MIME-Version: 1.0\n"
 "Content-Type: text/plain; charset=UTF-8\n"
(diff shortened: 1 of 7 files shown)


ifoche:
Task linked: [CU-869azwn9r Avoid home page refresh](https://app.clickup.com/t/869azwn9r)

gqcorneby (approved):
Thanks @deeonwuli! This fix worked :) It really looks like this fix is safe but the idea of using a long loading screen with just a timeout is very strange. 

Additional notes:
I tried checking a couple of functionalities and tried tracing why it was added in the first place (copied over from [training app](https://github.com/EyeSeeTea/training-app/pull/165/changes/f58707246087836a9eb903caa7e59a8cc140aece#diff-232989c0639e39ca5283f740bac316674cd6bce49703e7cc17ae3767b3b798e1)) and it doesn't seem like it was added for a specific functionality 🤔

gqcorneby:
Sorry for all the notifications on this, I went back and forth.

After checking a bit more, I think the long loading is there for actual initialization, like when the app is initializing the default modules. I don't think homepage app has default modules (unlike in training app) so I think removing the extra complexity of `isLoadingLong` is cleaner.

MiquelAdell (approved):

PR #67, opened 2026-05-05
Translate action description and import per-action

Ramon-Jimenez:
Closes: https://app.clickup.com/t/869d697pd 

## Summary
- Include action **description** in the JSON translation export/import flow (was silently limited to `name`, so `action-description` never appeared in the exported JSON and was ignored on import).
- Move **Import JSON translations** from the global toolbar to a per-row action (next to "Export JSON translations") and pre-select that row's action in the import dialog. The global entry is removed.

## Test plan
- [ ] Open an action with a non-empty description, run **Export JSON translations**: the JSON inside the ZIP contains both `action-name` and `action-description` keys.
- [ ] Translate the file, then on the same row click **Import JSON translations**, pick the language, confirm: both name and description are stored as translations and rendered in that locale.
- [ ] On a legacy action without a stored `description`, export still works and import does not throw.
- [ ] The global "Import JSON translations" button no longer appears; the per-row entry does.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/i18n/ar.po
+++ b/i18n/ar.po
@@ -1,7 +1,7 @@
 msgid ""
 msgstr ""
 "Project-Id-Version: i18next-conv\n"
-"POT-Creation-Date: 2026-02-05T16:47:42.188Z\n"
+"POT-Creation-Date: 2026-05-06T08:40:59.606Z\n"
 "PO-Revision-Date: 2025-06-06T19:59:41.289Z\n"
 "MIME-Version: 1.0\n"
 "Content-Type: text/plain; charset=UTF-8\n"
@@ -246,10 +246,10 @@ msgstr ""
 msgid "Export JSON translations"
 msgstr ""
 
-msgid "Import actions"
+msgid "Import JSON translations"
 msgstr ""
 
-msgid "Import JSON translations"
+msgid "Import actions"
 msgstr ""
 
 msgid "App/Link"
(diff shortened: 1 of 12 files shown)


gqcorneby (changes requested):
Hi @Ramon-Jimenez! This action selection in the import dialog always bothered me 😅  
Moving the action from global to each row makes more sense. Domain changes are fine, but I think we can clean up the component further and get rid of the action selection in this dialog entirely.

We can parallel the implementation in `LandingPageListTable` to “stash” the selected `actionId` in `ActionListTable` itself when the import row action is clicked, right before calling `startImport()` with no args.

That lets us drop:

1. The action picker `<Select>` in `ImportTranslationDialog`
2. `selectedAction` / `presetActionId` / `effectiveActionId` state
3. The `actions` / `translate` reads from `useAppContext`
4. The `type` prop (no longer needed to gate the picker)
5. The `key` argument from `onSave` and `startImport`

**Result:** `ImportTranslationDialog` becomes a generic “pick a language + upload JSON” component, reusable as-is across actions, landing pages, and notifications.

Each parent owns the key it cares about (`landingPageId`, `actionId`, or none for notifications).

Ramon-Jimenez:
@gqcorneby Please, proceed with the changes you are recommending at your convenience

gqcorneby:
Hi @Ramon-Jimenez! I applied the changes and tested. I don't know why I can't add you as reviewer :sweat_smile:

